PR #1, opened 2016-01-12
2016 01 mn add translatedfield

MagdaN:
I added a translatable Simple Page Model - migration is needed


Johnschmidfront:
Worked fine.

PR #3, opened 2016-01-14
add two-column-block

Johnschmidfront:
authors can add two-column-blocks on "Simple Pages". @MagdaN please review.


MagdaN:
@Johnschmidfront I am not sure if having a streamblock in a streamblock is working to well in the backend ui of wagtail - maybe we should split this into two blocks - one with an image or video on the left and the text on the right (Media Text Block) and one the other way around (Text Media Block )

Johnschmidfront:
@MagdaN Please review :)

MagdaN:
Works nicely

PR #5, opened 2016-01-19
2016 01 js carousel2

Johnschmidfront:
Not working - please don't merge. @MagdaN please fix it. thank you!


MagdaN:
@Johnschmidfront can you check an merge this please - you need to migrate. For this make sure that there are no other migration files than the one in the repo (last one 0003_simplepage.py ) and the one I added. The easiest is probably to delete your database and run the migrations.

MagdaN:
@Johnschmidfront can you test again

PR #20, opened 2016-01-29
Fixed highlight options

local-girl:
Fixed the highlighting options. Added a 'boxed' option (see homepage design "e-participation"


MagdaN:
Looks good!

PR #39, opened 2016-05-10
Added feedback button

localgrr:
Adds a simple feedback button


MagdaN:
Looks good, just one little thing, could you maybe disable (or make it a bit smaller) the button in mobile view

PR #41, opened 2016-06-03
Add process app to euth

slomo:


euth/process/util.py
```diff
@@ -0,0 +1,5 @@
+from django.conf.urls import url
+
+
+def processUrl(suburl, views, **kwargs):
+    return url(r'^(?P<process_name>\w+)/' + suburl, views, **kwargs)
```
MagdaN: This file is not used. Should it be removed?

euth/process/views.py
```diff
@@ -0,0 +1,26 @@
...
+    phases = process.phase_set.order_by('order')
+    context = {
+        'process': process,
+        'phases': phases,
+        'breadcrumbs': [
+            ('/', reverse('process-listing')),
+            (process.title, None),
+        ]
+    }
+    return render(request, 'process/detail.html', context)
```
MagdaN: Maybe this could be done with generic modelviews?

euth/process/models.py
```diff
@@ -0,0 +1,87 @@
...
+        ParticipationModule,
+        models.CASCADE,
+    )
+
+    date_start = models.DateTimeField(blank=True, null=True)
+    date_end = models.DateTimeField(blank=True, null=True)
+
+    def __str__(self):
+        return '{} phase of process «{}»'.format(
+            quantityStr(self.order), self.process.name)
```
MagdaN: We need to add a validation to the phase checking that there is no other overlapping phase.

slomo:
This code as been superseded by newer developments. Instead we will have pull-request for the projects app.

PR #42, opened 2016-06-03
Create comment with react and hyperscript

slomo:


MagdaN:
@slomo: I added some tests, do they look ok?

PR #49, opened 2016-06-15
2016 06 mn remove comments

MagdaN:


euth/comments/tests.py
```diff
@@ -99,3 +101,36 @@ def test_authenticated_user_can_reply_to_comment(self):
...
+
+    def test_admin_of_comment_can_set_censored_flag(self):
+        url = reverse('comments-detail', kwargs={'pk': self.comment1.pk})
+        self.client.force_authenticate(user=self.admin)
+        response = self.client.delete(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data['is_deleted'], True)
+        self.assertEqual(response.data['comment'], 'deleted by moderator')
+
+
```
slomo: OOOOOOH so manny empty lines :smile:

slomo:
Looks good. Can you please merge or rebase master?

slomo:
I  will continue on this branch

euth/comments/apps.py
```diff
@@ -1,5 +0,0 @@
-from django.apps import AppConfig
-
-
-class CommentsApiConfig(AppConfig):
-    name = 'comments'
```
keimlink: Why did you delete `apps.py`?

slomo: Because it seemed to that it wasn't used. As far as I see apps.py only is used if `__init.py__` does contains `default_app_config =`

keimlink: From Django's [Applications Documentation](https://docs.djangoproject.com/en/1.8/ref/applications/#configuring-applications):

> New applications should avoid default_app_config. Instead they should require the dotted path to the appropriate AppConfig subclass to be configured explicitly in INSTALLED_APPS.

This means:

1. An app should have an `apps.py` module
2. `INSTALLED_APPS` should use the dotted path to an `AppConfig` class in this module
3. An app should not use `default_app_config`, it only exists for backwards compatibility

PR #54, opened 2016-06-16
Test user management register and login

MagdaN:
@slomo : almost finished with the tests - just the test for activate is missing


euth/user_management/tests.py
```diff
@@ -1,6 +1,135 @@
...
+    testregistration = Registration.objects.create(
+        username='testuser2',
+        email='testuser2@liqd.de',
+        password='password',
+        token='b628de97-2a3a-4699-b4a4-16e64e4590f5',
+        next_action='/de/')
+    return testregistration
+
+
+@pytest.mark.django_db
```
slomo: Does this test really use the database?

euth/user_management/tests.py
```diff
@@ -1,6 +1,135 @@
...
+    register_url = reverse('register')
+    response = client.post(
+        register_url, {
+                'username': 'testuser2',
+                'email':'testuser@liqd.de',
+                'password': 'password',
+                'password_repeat': 'password'
+            }
+        )
+    assert Registration.objects.all().count() == 1
```
slomo: Maybe it should also be tested if the fields are correct (atleast username and email). It would also be interesting to reregister with the same credentials.

euth/user_management/tests.py
```diff
@@ -1,6 +1,135 @@
...
+
+@pytest.mark.django_db
+def test_activate_user(testregistration, client):
+    assert User.objects.all().count() == 0
+    token = testregistration.token
+    activate_url = reverse('activate', kwargs={'token': token})
+    client.post(activate_url, {})
+    #assert User.objects.all().count() == 1
+
+
```
slomo: Sooooo many empty lines.

MagdaN: Argh, sorry, I was not finished and just pushed my current state

euth/user_management/tests.py
MagdaN: yup

slomo:
@MagdaN @keimlink I rebased on master, fixed the comments I left and added some additional tests.

Also the views now return with status code 400 if form validation fails.

slomo:
I just realized that we do not test, whether emails have been send. So I will add this in an extra commit.

slomo:
Fixed it. We are now also testing the sending of emails.

euth/user_management/tests.py
```diff
@@ -1,6 +1,161 @@
...
+    logout_url = reverse('logout')
+    response = client.get(logout_url + '?next=/de/next_location')
+    assert response.status_code == 302
+    assert response.url == '/de/next_location'
+    assert '_auth_user_id' not in client.session
+
+
+@pytest.mark.django_db
+def test_register(client):
+    assert Registration.objects.all().count() == 0
```
MagdaN: Is this still needed?

euth/user_management/tests.py
```diff
@@ -1,6 +1,161 @@
...
+    response = client.post(register_url, data)
+    assert response.status_code == 200
+    data['email'] = 'anotheremail@liqd.de'
+    register_url = reverse('register')
+    response = client.post(register_url, data)
+    assert response.status_code == 400
+
+@pytest.mark.django_db
+def test_register_invalid(client):
+    assert Registration.objects.all().count() == 0
```
MagdaN: see above

euth/user_management/tests.py
```diff
@@ -1,6 +1,161 @@
...
+from django.contrib.auth.models import AnonymousUser, User, Permission
 from django.test import TestCase
 
-# Create your tests here.
-class DummyTestCase(TestCase):
-    def test_dummy(self):
-        self.assertTrue(True)
+from .forms import LoginForm
+from .models import Registration
+from .views import register_user
```
keimlink: It's easier to write and to read if you use `from . import views` instead of importing every view and model separately. 😄

keimlink:
A few comments on the general test layout:

1. It's common pratice in the Python and Django to separate code and tests into separate directories.
2. Pytest fixtures should be in a separate `conftest.py` module to separate them from the tests.

The file system structure could look like this:

```
euth_wagtail
├── Procfile
├── README.md
├── comments_api
├── contrib
├── euth
├── euth_wagtail
├── home
├── locale
├── manage.py
├── package.json
├── projects
├── pytest.ini
├── requirements.txt
├── runtime.txt
├── search
├── setup.py
└── tests
    ├── __init__.py
    ├── conftest.py
    └── user_management
        ├── __init__.py
        ├── conftest.py
        ├── factories.py
        ├── test_forms.py
        ├── test_managers.py
        ├── test_models.py
        └── test_views.py
```

- Fixtures that will be used by tests of other apps should be in `tests/conftest.py`, e.g. the `testuser` fixture.
- Fixtures that belong to the app should be in `tests/user_management/conftest.py`.
- Because of pytest's fixture loading mechanism test will still work as before.
- Imports in the test suite that reference code inside the project or apps must now be absolute: `from euth.user_management import models` instead `from . import models`

keimlink:
Instead of creating models manually in fixtures you can use [pytest-factoryboy](https://github.com/pytest-dev/pytest-factoryboy). It integrates [factory_boy](http://factoryboy.readthedocs.org/) with pytest.

Furthermore you can use [pytest-faker](https://github.com/pytest-dev/pytest-faker), a [faker](http://www.joke2k.net/faker/) integration for the pytest test runner. Faker generates random testing data.

factory_boy is also prepared to use faker, so you can use faker in two ways!

A simple example would look like this:

```
# tests/conftest.py

from pytest_factoryboy import register

from . import factories

register(factories.UserFactory)

# tests/factories.py
import factory


class UserFactory(factory.django.DjangoModelFactory):
    class Meta:
        model = 'euth.user_management.User'  # Equivalent to ``model = euth.user_management.models.User``
        django_get_or_create = ('username',)

    username = factory.Faker('name')
    password = 'password'

# tests/user_management/test_views.py

@pytest.mark.django_db
def test_login(client, user):
    login_url = reverse('login')
    response = client.get(login_url)
    assert response.status_code == 200
    response = client.post(
        login_url,
        {'username': user.username, 'password': user.password}
    )
    assert response.status_code == 302
    assert int(client.session['_auth_user_id']) == user.pk

def test_login_wrong_password(client, user, faker):
    login_url = reverse('login')
    response = client.post(
        login_url,
        {'username': user.username, 'password': faker.password()}
    )
    assert response.status_code == 400

```

If you need more ``User`` instances you can also use the ``user_factory``:

```
def test_factory_fixture(user_factory):
    user = user_factory(username="Ada")
    assert user.username == "Ada"
```

MagdaN:
I will move the tests to euth_wagtail/tests as @keimlink suggested in a seperate PR

PR #57, opened 2016-06-17
Add password reset functionality and html emails

slomo:


MagdaN:
@slomo I tested this, works nicely but I also found this, which could also be used: https://docs.djangoproject.com/en/1.8/topics/auth/default/#django.contrib.auth.views.password_reset

slomo:
Using the standard django views would probably be better. But since we want to user_managment with django-allauth, we should keep it for now.

PR #62, opened 2016-06-22
Switch to custom user model

slomo:
- set email as primary identifier
 - require email, username and password


slomo:
This changes the initial migration of `user_managment` requires the deletion of data_bases

MagdaN:
When I tried to register, I got: Manager isn't available; User has been swapped for 'user_management.User'

euth/user_management/models.py
```diff
@@ -1,5 +1,41 @@
...
+            'unique': _('A user with that email address already exists.'),
+       })
+    is_staff = models.BooleanField(_('staff status'), default=False,
+        help_text=_('Designates whether the user can log into this admin '
+                    'site.'))
+    is_active = models.BooleanField(_('active'), default=True,
+        help_text=_('Designates whether this user should be treated as '
+                    'active. Unselect this instead of deleting accounts.'))
+    date_joined = fields.AutoCreatedField
+
```
MagdaN: @slomo Should we also use model_utils.models.TimeStampedModel for users?

MagdaN:
Now it works!

euth/user_management/forms.py
```diff
@@ -47,7 +46,7 @@ def clean_password_repeat(self):
 
     def clean_username(self):
         username = self.cleaned_data.get('username')
-        user_exists = User.objects.filter(
+        user_exists = get_user_model().objects.filter(
```
keimlink: Instead of calling `get_user_model()` several times you can also place this line at the top of the module:

```
User = get_user_module()
```

This way all code using the `User` model stays the same. 😉

PR #68, opened 2016-06-23
Add organisations app

slomo:


euth/organisations/migrations/0001_initial.py
```diff
@@ -0,0 +1,36 @@
...
+            name='Organisation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, verbose_name='ID', primary_key=True, serialize=False)),
+                ('created', model_utils.fields.AutoCreatedField(verbose_name='created', editable=False, default=django.utils.timezone.now)),
+                ('modified', model_utils.fields.AutoLastModifiedField(verbose_name='modified', editable=False, default=django.utils.timezone.now)),
+                ('name', models.CharField(max_length=512, unique=True)),
+                ('slug', models.SlugField(max_length=512, unique=True)),
+                ('description_why', models.TextField()),
+                ('description_how', models.TextField()),
+                ('description', models.TextField()),
```
MagdaN: I think all description fields should have translations

euth/organisations/migrations/0001_initial.py
```diff
@@ -0,0 +1,36 @@
...
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Organisation',
+            fields=[
+                ('id', models.AutoField(auto_created=True, verbose_name='ID', primary_key=True, serialize=False)),
```
MagdaN: Ids are created for all models automatically by django

euth/organisations/models.py
```diff
@@ -0,0 +1,28 @@
...
+    def get_by_natural_key(self, name):
+        return self.get(name=name)
+
+
+class Organisation(model_utils.TimeStampedModel):
+    name = models.CharField(max_length=512, unique=True)
+    slug = models.SlugField(max_length=512, unique=True)
+    description_why = models.TextField()
+    description_how = models.TextField()
+    description = models.TextField()
```
MagdaN: I think all descriptions need translations

slomo:
I propose to add the translated fields in a later pull-request. There are some interesting questions: 

- should the name also be translated
- from which language should the slug be derived

MagdaN:
As english is the default language (which also is required I think) I would derive the slug from the english title

euth/organisations/models.py
keimlink: No, we are planning to use [django-parler](https://github.com/django-parler/django-parler) for a different project. The main difference is how the translations are stored: django-modeltranslation adds fields for each language to the model while django-parler uses a separate table.

PR #98, opened 2016-07-04
2016 06 cc orgs page style

localgrr:
Some styling of the org page, not much else to do until projects and processes are listed


MagdaN:
looks good, I will merge so PM and Design can check on heroku

PR #100, opened 2016-07-05
Add validators for organisation images

slomo:
Implement validators for organisation image

### DIFF
--- a/euth/contrib/validators.py
+++ b/euth/contrib/validators.py
@@ -1,0 +1,23 @@
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
+
+def validate_image(image, min_width, min_height):
+    errors = []
+    if image.width < min_width:
+        msg = _('Image must be at least {min_width} pixels wide')
+        errors += ValidationError(msg.format(min_width=min_width))
+    if image.height < min_height:
+        msg = _('Image must be at least {min_height} pixels high')
+        errors += ValidationError(msg.format(min_height=min_height))
+    if errors:
+        raise ValidationError(errors)
+    return image
+
+
+def validate_hero_image(image):
+    validate_image(image, 1300, 600)
+
+
+def validate_logo(image):
+    validate_image(image, 400, 400)
--- a/euth/contrib/validators.py
+++ b/euth/contrib/validators.py
@@ -1,28 +1,8 @@
-from django.core.exceptions import ValidationError
-from django.utils.translation import ugettext_lazy as _
-
-image_max_mb = 5
-
-def validate_image(image, min_width, min_height):
-    errors = []
-    image_max_size = image_max_mb * 10**6
-    if image.size > image_max_size:
-        msg = _('Image should be at most {max_size} MB')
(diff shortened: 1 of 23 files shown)


slomo:
@keimlink Do I need to do some further validation of the file type (like python-magic)? As far as I see Django image field will already handle this.

Also is there a convenient library to generate text pictures? Or should I use pillow / willow for this?

euth/organisations/models.py
```diff
@@ -1,7 +1,9 @@
 from django.conf import settings
 from django.db import models
+
 from model_utils import models as model_utils
 
+from ..contrib import validators
```
keimlink: I would advise to use relative imports only inside an app. Otherwise absolute imports are better; in this case `from euth.contrib import validators`. It makes the apps independent from the overall project structure which simplifies a future extraction of the app.

euth/contrib/validators.py
```diff
@@ -0,0 +1,28 @@
...
+        errors += ValidationError(msg.format(min_width=min_width))
+    if image.height < min_height:
+        msg = _('Image must be at least {min_height} pixels high')
+        errors += ValidationError(msg.format(min_height=min_height))
+    if errors:
+        raise ValidationError(errors)
+    return image
+
+
+def validate_hero_image(image):
```
keimlink: One could argue that only `validate_image()` should be in `euth.contrib`, because the other two functions do the validation of specific image type. So these two could be moved to `euth.organisations.validators`.

euth/contrib/validators.py
```diff
@@ -0,0 +1,28 @@
...
+from django.utils.translation import ugettext_lazy as _
+
+image_max_mb = 5
+
+def validate_image(image, min_width, min_height):
+    errors = []
+    image_max_size = image_max_mb * 10**6
+    if image.size > image_max_size:
+        msg = _('Image should be at most {max_size} MB')
+        errors += ValidationError(msg.format(max_size=image_max_mb))
```
keimlink: You can't append to a `list` using the `+=` operator. Unfortunately this does not result in an error but in an empty list.

keimlink:
@slomo Yes, Django does basic validation if the file is a valid image. At @transcode-de we use Pillow in a pytest fixture to create images for testing. If you're actually creating an organisation with an image it's helpful to change the `MEDIA_ROOT` setting to a temporary location to prevent pollution of your development `MEDIA_ROOT` with test files.

euth/contrib/validators.py
slomo: Interestingly this works nevertheless, I guess ValidationError seems to support the `+` anyway.

euth/contrib/validators.py
slomo: I'd like to use the hero_image_validator also in the projects app. Would you recommend to still move the logo validator to the organisations app?

MagdaN: You could add hero_image-validation and logo_validation in the base dir of the euth apps. Could that be an option?

keimlink: If you want to reuse the validators you should keep them in `euth.contrib`. 😃

keimlink:
LGTM 👍

PR #101, opened 2016-07-05
add autofixtures with faker

MagdaN:
@keimlink : I added the autofixtures, now I would like that they are automatically added on heroku, but only once. I thought of adding management scripts to each app that checks if a given number of objects, e.g. users, still exists and only if that is not the case add the autofixtures. Something like:

`python manage.py loaduserautofixtures 20`

and within the command code

`if User.objects.all().count() < number:`
        `call_command('loadtestdata', 'user_management.User:' + number)`

But this seems kind of complicated. Do you have an idea?


keimlink:
I would only automate this if the task is executed very often. Otherwise creating (and testing) the commands takes more time then to run them manually. I assume that the test data must only be generated if the staging server has been reset, which shouldn't happen too often.

euth/comments/autofixtures.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib.contenttypes.models import ContentType
+
+from .models import Comment
+from home.models import HomePage
+from autofixture import generators, register, AutoFixture
+
+class CommentAutoFixture(AutoFixture):
+
+    homepage_contenttype = ContentType.objects.get(app_label='home', model='homepage')
+    homepage_id = HomePage.objects.all().first().pk
```
keimlink: If you use `first()` you don't have to use `all()`.

euth/organisations/autofixtures.py
```diff
@@ -0,0 +1,14 @@
+from .models import Organisation
+from autofixture import register, AutoFixture
+from faker import Faker
+from .. import generators
+
+
+class OrganisationAutoFixture(AutoFixture):
+
+    field_values = {
+        'name': generators.FakeOrganisationNameGenerator(),
```
keimlink: It's not wrong to implement a `FakeOrganisationNameGenerator` class or other classes. But you can also use `autofixture.generators.CallableGenerator()` and call `faker` directly like so:

```python
from autofixture import AutoFixture, generators, register
from faker import Factory

from . import models

fake = Factory.create()


class OrganisationAutoFixture(AutoFixture):

    field_values = {
        'name': generators.CallableGenerator(fake.company),
        'slug': generators.CallableGenerator(fake.slug),
    }


register(models.Organisation, OrganisationAutoFixture)
```

This reduces the amount of code you have to maintain and that can break. 😃

MagdaN: Ah, ok thats better, I will use it as you suggested. Thanks.

euth/comments/autofixtures.py
```diff
@@ -0,0 +1,19 @@
...
+from autofixture import generators, register, AutoFixture
+
+class CommentAutoFixture(AutoFixture):
+
+    homepage_contenttype = ContentType.objects.get(app_label='home', model='homepage')
+    homepage_id = HomePage.objects.first().pk
+
+    field_values = {
+        'content_type': homepage_contenttype,
+        'object_pk': homepage_id,
```
slomo: @keimlink Is there a good way to build a auto-fixture for generic forgein keys? I imagine a function that takes a list of models and will automatically select a random instance from those models and append a comment to it. 

The function seems trivial but how do I set fixture values that are somehow connected?

slomo:
Looks promising

euth/comments/autofixtures.py
keimlink: @slomo What do you mean by "fixture values that are somehow connected"? You could use the [AutoFixture.post_process_instance() hook](http://django-autofixture.readthedocs.io/en/latest/usage.html#autofixture.base.AutoFixture.post_process_instance) to add generic foreign keys. Does this answer your question?

slomo: Yeah probably. I will give it a try.

PR #103, opened 2016-07-05
2016 07 mn translations

MagdaN:
fixes https://github.com/liqd/euth_wagtail/issues/99

### DIFF
--- a/home/templatetags/url_translations.py
+++ b/home/templatetags/url_translations.py
@@ -5,10 +5,7 @@ from django.http import Http404
 from django.utils import translation
 
 
-
 register = template.Library()
-
-
 
 
 @register.simple_tag(takes_context=True)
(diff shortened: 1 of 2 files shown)


home/templatetags/url_translations.py
```diff
@@ -4,17 +4,35 @@
 from django.http import Http404
 from django.utils import translation
 
+
+
```
slomo: Why so many spaces?

home/templatetags/url_translations.py
```diff
@@ -4,17 +4,35 @@
...
-		if view.args:
-			url = '/' + language + '/' + view.args[0]
-		else:
-			url = '/'  + language + '/'
-	except Http404:
-		url = '/'  + language + '/'
-	return url
+
+
+@register.simple_tag(takes_context=True)
```
slomo: even more spaces!

home/templatetags/url_translations.py
```diff
@@ -4,17 +4,35 @@
...
+            if url_parts.args:
+                url = '/' + lang + '/' + url_parts.args[0]
+            else:
+                url = '/'  + lang + '/'
+        finally:
+            translation.activate(cur_language)
+    except Http404:
+        url = '/'  + lang + '/'
+
+    return "%s" % url
```
slomo: This could porbably also be `str(url)`, not sure what is better.

home/templatetags/url_translations.py
```diff
@@ -4,17 +4,35 @@
...
+            url = reverse(url_parts.view_name, kwargs=url_parts.kwargs)
+        except:
+            if url_parts.args:
+                url = '/' + lang + '/' + url_parts.args[0]
+            else:
+                url = '/'  + lang + '/'
+        finally:
+            translation.activate(cur_language)
+    except Http404:
+        url = '/'  + lang + '/'
```
slomo: Is this case ever reached? Also the logic looks very similar to the else in the other except.

PR #106, opened 2016-07-07
2016 07 cc style projects

localgrr:
Still working on this


localgrr:
Thats as much as I can do until we have projects to style - will do that in a further request

PR #115, opened 2016-07-08
2016 07 mn display projects for organisations

MagdaN:


euth_wagtail/settings/base.py
```diff
@@ -150,7 +150,8 @@
 
 THUMBNAIL_ALIASES = {
     '': {
-        'organisation_thumbnail': {'size': (80, 80), 'crop': True }
+        'organisation_thumbnail': {'size': (100, 100), 'crop': True },
```
slomo: May we should rather have `'crop':  'smart'` here. See [easy-thumbnail docs](http://easy-thumbnails.readthedocs.io/en/latest/usage/#thumbnail-options).

euth_wagtail/templates/base.html
```diff
@@ -55,12 +52,16 @@
         </div>
 
         {% block footer %}
-
+        {% include "includes/footer.html" %}
```
slomo: Why wasn't this needed before?

MagdaN: Befor the footer was added on each page, because on some pages (where A3 is embedded) the footer should not show

PR #117, opened 2016-07-11
Js translations with django translation view

slomo:
- also change were static files are looked up, because otherwise collection takes to much time


MagdaN:
LGTM

PR #118, opened 2016-07-11
2016 07 mn add image typecheck to validtion

MagdaN:


tests/conftest.py
```diff
@@ -24,3 +24,13 @@ def smallImage():
...
     return factory.django.ImageField(width=1400, height=1400)
+
+
+@pytest.fixture
+def ImageBMP():
+    return factory.django.ImageField(width=1400, height=1400, format='BMP')
+
+
+@pytest.fixture
+def ImagePNG():
```
slomo: What is the default format of that factory?

MagdaN: JPEG

PR #121, opened 2016-07-12
add heroimage thumbnail

MagdaN:


MagdaN:
@localgrr can you please review?

PR #123, opened 2016-07-13
2016 06 cc projects carousel

localgrr:
Added carousel to projects page + a plugin. The example in project_detail.html is just an example though, this should be overwritten by real data.


MagdaN:
@localgrr did you forget to commit _carousel.scsss

MagdaN:
@localgrr can you change the js?

MagdaN:
Is this supposed to look like that? Or should the tiles on the right- and the lefthand side of the arrows disappear? 
bildschirmfoto 2016-07-19 um 10 42 17

NadineKarbach:
@MagdaN The titles, actually the whole box, left and right the arrow should disappear in the sense of being blurred white/shadow white

localgrr:
@NadineKarbach that blurring function is a bit cryptic, I'd prefer to keep it simple for now

localgrr:
OK I made that blur out thing, i still think there should be a forward and back navigation though ? @nicoladimarco ?

nicoladimarco:
@localgrr @MagdaN  the best would be if the arrows would appear on top of the semi transparent card (both left and right )

localgrr:
I took away the faded opacity option on mobile, tbh I think we should remove it in general it makes the interface a bit clunky imho

PR #127, opened 2016-07-15
Make project detail configurable by phase

slomo:


euth/phases/contents.py
```diff
@@ -0,0 +1,35 @@
...
+    def __iter__(self):
+        return [  '{}:{}:{}'.format(module, values.weight, phase)
+           for (module, phases) in self._registry.items()
+           for phase, values in phases.items() ].__iter__()
+
+
+    def choices(self):
+        return [ ('{}:{}:{}'.format(module, values.weight, phase), '{} {}'.format(module, phase))
+           for (module, phases) in self._registry.items()
+           for phase, values in phases.items() ]
```
MagdaN: Whats happening here?

MagdaN:
LGTM but maybe needs some comments

euth/phases/contents.py
```diff
@@ -0,0 +1,30 @@
+from django.apps import apps
+
+class PhaseContent():
+    @property
+    def identifier(self):
+        return '{}:{:03d}:{}'.format(self.app, self.weight, self.phase)
```
keimlink: You can also just pass `self` into `format()` and define the position of the values inside the string:
```python
'{s.app}:{s.weight:03d}:{s.phase}'.format(s=self)
```
I like this approach because it's flexible and does not depend on the argument ordering. 😃

euth/phases/__init__.py
```diff
@@ -0,0 +1 @@
+from .contents import content
```
keimlink: Why not also importing `PhaseContent` so that apps using phase need only a single import?

euth/phases/admin.py
```diff
@@ -1,5 +1,20 @@
...
+class PhaseForm(forms.ModelForm):
+    type = forms.ChoiceField(choices=content.as_choices)
+
+    class Meta:
+        fields = '__all__'
+        model = models.Phase
+
+
+class PhaseAdmin(admin.ModelAdmin):
+    model = models.Phase
```
keimlink: AFAIK there is no `ModelAdmin.model` attribute.

euth/phases/apps.py
```diff
@@ -1,6 +1,11 @@
...
 
+from . import content
 
 class PhasesConfig(AppConfig):
     name = 'euth.phases'
     label = 'euth_phases'
+
+    def ready(self):
+        from django.utils.module_loading import autodiscover_modules
+        autodiscover_modules('phases', register_to=content)
```
keimlink: You can also use `self.module.content` instead of `content` (see [AppConfig.module](https://docs.djangoproject.com/en/1.8/ref/applications/#django.apps.AppConfig.module)).

euth/phases/contents.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __str__(self):
+        return '{}({}:{})'.format(self.__class__.__name__, self.app, self.phase)
+
+
+class PhaseContents():
+    def __init__(self):
+        self._registry = {}
```
keimlink: If you don't do anything with `self._registry` in `__init__()`  you can simply set the default by defining an attribute:
```python
class PhaseContents():
    _registry = {}

    def __getitem__(self, identifier):
    ...
```

euth/phases/contents.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __getitem__(self, identifier):
+        if type(identifier) != str:
+            raise TypeError('Phase identifier must be str')
+        self._registry[identifier]
```
keimlink: Shouldn't ` __getitem__()` return `self._registry[identifier]`? 😉

euth/phases/models.py
```diff
@@ -1,15 +1,29 @@
 from django.conf import settings
 from django.db import models
-from model_utils import models as model_utils
 
 from euth.modules import models as modules_models
+from .validators import validate_content
+from . import content
+
+class PhasesManager(models.Manager):
```
keimlink: To be more flexible you should use a `QuerySet` instead of a `Manager`. `QuerySet` calls can be chained, what you can't do with a `Manager`.

This works only with a `QuerySet`:
```python
Phase.objects.all_phases().filter(description__icontains="foo").active_phase()
```
If you use a `Manager` the `QuerySet` created by `filter()` will not have the `active_phase()` method.

The custom `QuerySet` would look like this:

```python
class PhaseQuerySet(models.QuerySet):

    def active_phase(self, project):
        return self.filter(module__project=project).order_by('type').first()

    def all_phases(self, project):
        return self.filter(module__project=project).order_by('type')
```

And will be added to the model as a manager as follows:

```python
class Phase(models.Model):

    # fields

    objects = PhaseQuerySet.as_manager()
```

For more details see the [Creating a manager with QuerySet methods](https://docs.djangoproject.com/en/1.9/topics/db/managers/#creating-a-manager-with-queryset-methods).

euth/projects/mixins.py
```diff
@@ -0,0 +1,23 @@
...
+        self.project = models.Project.objects.get(slug=kwargs['project'])
+        self.phase = phases_models.Phase.objects.active_phase(self.project)
+        self.module = self.phase.module if self.phase else None
+        return super(ProjectMixin, self).dispatch(*args, **kwargs)
+
+    def get_context_data(self, *args, **kwargs):
+        context = super(ProjectMixin, self).get_context_data(*args, **kwargs)
+        context['project'] = self.project
+        context['phase'] = self.phase
+        context['module'] = self.module
```
keimlink: You can simply use `view.project` or `view.phase` in a template, because `django.views.generic.base.ContextMixin` adds the current view to the context. This makes adding class attributes to the context superfluous.

If you are not sure if all views will use `django.views.generic.base.ContextMixin` you can simply extend it with your mixin.

keimlink:
Rest of the code LGTM. Great idea to model phases. 👍

PR #130, opened 2016-07-15
add basic styling for idea detail page

MagdaN:


MagdaN:
@localgrr there should be style in zeplin

PR #131, opened 2016-07-15
2016 07 mn add ck editor

MagdaN:


slomo:
Please add a validator that whitelists tags and attributes using bleach.

MagdaN:
@keimlink the relevant commit is this one: https://github.com/liqd/euth_wagtail/pull/131/commits/6e0aa7971f05632e23d2cc7e0c1685ca68389418

euth/ideas/models.py
```diff
@@ -1,3 +1,5 @@
+import bleach
```
slomo: unused import?

euth/ideas/models.py
```diff
@@ -6,6 +8,7 @@
 from django.utils.functional import cached_property
 
 from euth.contrib import validators
+from contrib.transforms import html_transforms
```
slomo: Where does this import comes from?

euth/ideas/models.py
```diff
@@ -22,6 +25,10 @@ def get_absolute_url(self):
         from django.core.urlresolvers import reverse
         return reverse('idea-detail', args=[str(self.slug)])
 
+    def clean(self):
```
slomo: Shouldn't his also call `super().clean()` ?

MagdaN: I used this code as example: https://docs.djangoproject.com/es/1.9/ref/models/instances/#django.db.models.Model.clean

euth/ideas/models.py
```diff
@@ -20,6 +22,11 @@ def get_absolute_url(self):
...
+        super().clean()
```
keimlink: If you [take a look at the source code](https://docs.djangoproject.com/en/1.8/_modules/django/db/models/base/#Model.clean) you will recognize that `clean()` is just an empty hook. So no need to call `super()` here. 😄

euth/ideas/urls.py
```diff
@@ -1,10 +1,11 @@
 from django.conf.urls import url
+from django.contrib.auth.decorators import login_required
 
 from . import views
 
 urlpatterns = [
     url(r'create/module/(?P<slug>[-\w_]+)/$',
-        views.IdeaCreateView.as_view(), name='idea-create'),
+        login_required(views.IdeaCreateView.as_view()), name='idea-create'),
```
keimlink: Instead of wrapping the view in URLconf you can use [django-braces' `LoginRequiredMixin`](https://django-braces.readthedocs.io/en/latest/access.html#loginrequiredmixin) to require a login. This is useful if you reuse the view somewhere else because the view itself will require the login and not just the instance used in URLconf.

euth/ideas/models.py
```diff
@@ -20,6 +22,11 @@ def get_absolute_url(self):
...
+        self.description = html_transforms.clean_html_field(
+            self.description)
```
keimlink: Of course it is right to do the transformation of the description in `Model.clean()`. But you have to remember that `Model.clean()` is not called if you just call `Model.save()` (see [Validating objects](https://docs.djangoproject.com/en/1.8/ref/models/instances/#validating-objects)). Only a `ModelForm` will call `Model.clean()`. So if you want to make sure that the description is always transformed you should move the code to `Model.save()`.

keimlink:
@MagdaN Even if this branch has already been merged I added three comments. Maybe this is something that helps you to improve the code in a future pull request. 😃

euth/ideas/models.py
MagdaN: I thought maybe we will implement clean() in the Item at some time and then we wonder why it does not work.

MagdaN:
@keimlink thanks! I will add your suggestions in a separate PR!

euth/ideas/models.py
keimlink: Sounds reasonable. 😉

PR #132, opened 2016-07-15
Add Makefile for development setup

slomo:


MagdaN:
Works nicely except for the trap command

PR #134, opened 2016-07-15
Replace model utils by custom base models

slomo:


euth/contrib/base_models.py
```diff
@@ -0,0 +1,15 @@
...
+class TimeStampedModel(models.Model):
+
+    created = models.DateTimeField(editable=False, default=timezone.now)
+    modified = models.DateTimeField(blank=True, null=True, editable=False)
+
+    class Meta:
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        if not self.pk is None:
```
keimlink: This expression is hard to read and understand. I would go for one of these:

1. `if self.pk`
2. `if self.pk is not None`

keimlink:
Except the single comment everything else looks god to me. 👍

tests/contrib/test_base_models.py
```diff
@@ -0,0 +1,46 @@
...
+from django.core.management.color import no_style
+from django.utils import timezone
+
+import freezegun
+import pytest
+
+from euth.contrib import base_models
+
+@pytest.fixture
+def time_stamped_model(request, db):
```
keimlink: I would suggest to put fixtures into ` tests/contrib/conftest.py` so that they are separated from the tests and can be reused by other tests.

tests/contrib/test_base_models.py
```diff
@@ -0,0 +1,46 @@
...
+    request.addfinalizer(fin)
+    return model
+
+def test_model(time_stamped_model):
+    TestModel = time_stamped_model
+
+    with freezegun.freeze_time("2016-06-11"):
+        obj = TestModel.objects.create()
+        assert obj.created == timezone.now()
+        created = timezone.now()
```
keimlink: What's the purpose of this line?

euth/contrib/base_models.py
slomo: A classical case of Yoda speech it is.

keimlink:
I created the branch [improve-timestampedmodel-tests](https://github.com/liqd/euth_wagtail/tree/improve-timestampedmodel-tests) with a different approach. Main difference is that I use a "test app" to create the model to test the abstract model instead of doing it in a fixture by using `django.db.connection`.

slomo:
The advantage being that this approach is less hacky? Because otherwise it looks like a disadvantage to me to spread the code into to more different modules.

keimlink:
Exactly. I can imagine that the code in `derive_mixin()` is hard to understand for people that didn't write it. And maybe it's even hard to understand for yourself after a few weeks?

The code in the test app uses only public APIs that are safe to use, so the possibility that this code will break with future updates is very low.

The new test app can also be used for other purposes, for example if you need a view or something else it can be added to this app.

slomo:
replaced by #157

PR #145, opened 2016-07-20
2016 07 mn delete thumbnails

MagdaN:
with this also easy thumbnail images are deleted after the deletion of a model instance


euth/projects/signals.py
```diff
@@ -0,0 +1,12 @@
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+from easy_thumbnails.files import get_thumbnailer
+
+from .models import Project
+
+
+@receiver(post_delete, sender=Project)
+def delete_images_for_Project(sender, instance, **kwargs):
```
slomo: Doesn't this naming break the function naming convention?

euth/organisations/signals.py
```diff
@@ -0,0 +1,15 @@
+from django.db.models.signals import post_delete
+from django.dispatch import receiver
+from easy_thumbnails.files import get_thumbnailer
+
+from .models import Organisation
+
+
+@receiver(post_delete, sender=Organisation)
+def delete_images_for_Organusation(sender, instance, **kwargs):
```
slomo: Same as below and typo in function name

MagdaN:
@keimlink do you think it would be better to use something like https://github.com/akolpakov/django-unused-media (the problem might be that this would also delete the thumbnails)

keimlink:
Your solution looks good. 👍

PR #152, opened 2016-07-21
2016 07 mn better ck editor

MagdaN:
this implements the suggestions made in  this PR https://github.com/liqd/euth_wagtail/pull/131

### DIFF
--- a/euth/ideas/models.py
+++ b/euth/ideas/models.py
@@ -18,14 +18,14 @@ class Idea(module_models.Item):
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs):
+        self.description = html_transforms.clean_html_field(
+            self.description)
+        super(Idea, self).save(*args, **kwargs)
+
     def get_absolute_url(self):
         from django.core.urlresolvers import reverse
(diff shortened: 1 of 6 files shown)


euth/ideas/models.py
```diff
@@ -18,15 +18,15 @@ class Idea(module_models.Item):
...
+
     def get_absolute_url(self):
         from django.core.urlresolvers import reverse
         return reverse('idea-detail', args=[str(self.slug)])
 
-    def clean(self):
-        super().clean()
-        self.description = html_transforms.clean_html_field(
-            self.description)
-
```
slomo: Why the move from save() to clean()?

MagdaN: Quote from @keimlink:
> Of course it is right to do the transformation of the description in Model.clean(). But you have to remember that Model.clean() is not called if you just call Model.save() (see Validating objects). Only a ModelForm will call Model.clean(). So if you want to make sure that the description is always transformed you should move the code to Model.save().

slomo: Ah okay. Good point.

PR #158, opened 2016-07-22
update comments styling

MagdaN:


MagdaN:
@localgrr can you take a look at it, and finalize it?

PR #162, opened 2016-07-26
2016 07 mn add idea list

MagdaN:
@localgrr can you finalize this?


localgrr:
I had some login troubles so worked on #163 instead

PR #168, opened 2016-08-03
Improve image upload widget

localgrr:
Better file uploading

### DIFF
--- a/euth/contrib/widgets.py
+++ b/euth/contrib/widgets.py
@@ -11,7 +11,7 @@ class ImageInputWidget(widgets.ClearableFileInput):
             .render(name, value, attrs)
 
         substitutions = {
-            'filename': basename(value.name),
+            'filename': '',
             'file_placeholder': ugettext(
                 'Select a picture from your local folder.'
             ),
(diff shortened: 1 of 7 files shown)


MagdaN:
Looks good on create form but I am using the same form also for updating, and here it looks not so good:
bildschirmfoto 2016-08-04 um 10 09 09

MagdaN:
Errors used to look like that, can you fix this please: 
bildschirmfoto 2016-08-04 um 10 14 34

localgrr:
Hi @MagdaN  can you expand out the upload field template tag so I can have more control over the HTML please ? thanks

slomo:
Okay, what is going on here? This pull-request is open since quite some time. Can we get it flying again?

MagdaN:
@slomo we need to figure out how to style the 'change image field'

slomo:
I have added a custom widget for the form, by overriding some values of the original widget. We could now continue to override even more (changing the html structure completely), but that would be not that trivial. Maybe either @localgrr  or @vellip can give me hint, what we would actually require here in terms of html.

slomo:
This looks very odd. I would say we should try to rebase this, rather than merging because now it contains  over 2000 lines of code ;)

slomo:
Fixed by rebase and force push

PR #182, opened 2016-08-11
 Add additional phases to idea collection

slomo:
This pull-request does not ensure that votingd/commenting is disabled on Idea detail pages. Because it doesn't deal with detail pages yet.

### DIFF
--- a/euth/comments/static/comments/react_comments.js
+++ b/euth/comments/static/comments/react_comments.js
@@ -133,7 +133,7 @@ var CommentBox = React.createClass({
       h('div.black-divider',
         this.state.comments.length + ' ' + django.ngettext('comment', 'comments', this.state.comments.length)),
       h('div.commentBox', [
-        h(CommentForm, {
+        this.props.isReadOnly ? null : h(CommentForm, {
           subjectType: this.props.subjectType,
           subjectId: this.props.subjectId,
           onCommentSubmit: this.handleCommentSubmit,
@@ -145,7 +145,8 @@ var CommentBox = React.createClass({
             comments: this.state.comments,
             handleCommentDelete: this.handleCommentDelete,
             handleCommentSubmit: this.handleCommentSubmit,
-            handleCommentModify: this.handleCommentModify
+            handleCommentModify: this.handleCommentModify,
+            isReadOnly: this.props.isReadOnly
           })
         ])
       ])
@@ -186,7 +187,8 @@ var CommentList = React.createClass({
           positiveRates: comment.rates.positive_rates,
           negativeRates: comment.rates.negative_rates,
           userRate: comment.rates.current_user_rate_value,
-          userRateId: comment.rates.current_user_rate_id
+          userRateId: comment.rates.current_user_rate_id,
+          isReadOnly: this.props.isReadOnly
         },
           comment.comment
         )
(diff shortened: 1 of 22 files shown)


euth/comments/static/comments/react_comments.js
```diff
@@ -221,11 +223,7 @@ var Comment = React.createClass({
   },
 
   allowForm: function () {
-    return !(this.props.content_type === this.context.comments_contenttype)
-  },
-
-  allowRate: function () {
-    return !(this.state.is_deleted)
+    return !this.props.isReadOnly && !(this.props.content_type === this.context.comments_contenttype)
```
keimlink: Why not use `!==` instead of `!( … === …)`?

euth/ideas/templates/euth_ideas/idea_detail.html
```diff
@@ -10,7 +10,7 @@
                     <a href="{% url 'project-detail' idea.project.slug %}"><i class="fa fa-arrow-left" aria-hidden="true"></i> {% trans 'back to project' %}</a>
                 </li>
             </ul>
-            {% if request.user == idea.creator or request.user.is_superuser %}
+            {% if request.user == idea.creator and view.crud_enabled or request.user.is_superuser %}
```
keimlink: What about using [django-rules](https://github.com/dfunckt/django-rules#rules-and-permissions-in-templates) instead?

euth/modules/mixins.py
```diff
@@ -0,0 +1,17 @@
...
+    def comment_enabled(self):
+        return self._feature_enabled('comment')
+
+    @property
+    def crud_enabled(self):
+        return self._feature_enabled('crud')
+
+    @property
+    def rate_enabled(self):
+        return self._feature_enabled('rate')
```
keimlink: You could also replace all properties with `__getattr__()` if want a more generic approach. 😃

euth/projects/models.py
```diff
@@ -53,3 +53,8 @@ def other_projects(self):
     @functional.cached_property
     def is_private(self):
         return not self.is_public
+
+    @functional.cached_property
+    def active_phase(self):
+        module = self.module_set.first()
+        return module.phase_set.order_by('type').first()
```
keimlink: Why not add a `related_name` to `Phase.module`?

tests/phases/test_contents.py
```diff
@@ -19,3 +20,16 @@ class DummyPhase(PhaseContent):
...
+
+
+def test_phase_features():
+    class DummyPhase(PhaseContent):
+        app = 'dummy'
+        phase = 'phase'
+        weight = 9999
+        features = {
+            'comment': (models.Post,)
+        }
```
keimlink: Because you are reusing the dummy from `test_phase_content()` I would suggest to add a factory or at least a fixture to create the dummy.

tests/projects/test_mixins.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+from django.views.generic import ListView
+from tests.apps.blog import models
```
keimlink: I would suggest to use a relative import. `apps.blog` is also part of the test suite.

euth/ideas/templates/euth_ideas/idea_detail.html
slomo: I'll add rules in another pull-request and will than adapt those checks.

PR #183, opened 2016-08-12
2016 08 mn report comments

MagdaN:
@keimlink @slomo I am not too sure about the post_save signal that sends emails to all admins telling them that an object has been reported. The problem is, that I don't have access to the request and therefore cannot create a full url, as the protocol is missing (problem described here: http://stackoverflow.com/questions/15520246/django-signal-get-request-full-path) 

This is why I added the protocol to the settings which I guess is not so nice. Alternatively, I could move the email-sending-code to the view, but then, if reports are not created via the API we need to add the code for sending emails again at that point. What do you think?


euth/reports/signals.py
```diff
@@ -0,0 +1,54 @@
...
+        try:
+            admin_url = urlresolvers.reverse("admin:%s_%s_change" %
+                                             (obj._meta.app_label,
+                                              obj._meta.model_name),
+                                             args=(obj.pk,))
+        except:
+            admin_url = ''
+
+        recievers = []
+        for user in get_user_model().objects.filter(is_superuser=True):
```
slomo: [ u.email for u in get_user_model().objects.filter(is_superuser=True) ]

MagdaN:
@keimlink I updated the PR it is now ready for review.

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+
+def get_creator_field(obj):
+    fields = obj._meta.fields
+
+    for f in fields:
+        field_object, model, direct, m2m = obj._meta.get_field_by_name(f.name)
+        if not m2m and direct and isinstance(field_object, ForeignKey):
+            if field_object.rel.to == get_user_model():
+                return f.name
+
```
MagdaN: this function could be much more easier if we would consider this: https://github.com/liqd/euth_wagtail/issues/184

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+        name = obj._meta.verbose_name.title()
+
+        try:
+            admin_url = urlresolvers.reverse("admin:%s_%s_change" %
+                                             (obj._meta.app_label,
+                                              obj._meta.model_name),
+                                             args=(obj.pk,))
+            admin_url = request.build_absolute_uri(
+                admin_url)
+        except:
```
keimlink: Using `except` without any expression is not good. This can hide exceptions you didn't intend to. What exactly do you want to catch here?

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+            admin_url = ''
+
+        moderators = get_user_model().objects.filter(is_superuser=True)
```
keimlink: Instead of using `get_user_model()` several times you can put this line at the top of the module (for example after the imports):
```python
User = get_user_model()
```
This allows you to write code like this:
```python
User.objects.filter(is_superuser=True)
```

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+
+def send_email_to_moderators(request, report):
+
+    obj = report.content_object
+
+    if obj:
+        name = obj._meta.verbose_name.title()
+
+        try:
+            admin_url = urlresolvers.reverse("admin:%s_%s_change" %
```
keimlink: I would suggest to use `format()` instead of `%`. This would allow something like this:
```python
"admin:{m.app_label}_{m.model_name}_change".format(m=obj._meta)
```
IMO `format()` has a syntax that is easier to read and understand.

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+                                             args=(obj.pk,))
+            admin_url = request.build_absolute_uri(
+                admin_url)
+        except:
+            admin_url = ''
+
+        moderators = get_user_model().objects.filter(is_superuser=True)
+
+        recievers = [
+            u.email for u in moderators]
```
keimlink: You don't need a line break here. 😉

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+
+        context = Context({
+            'site': Site.objects.get_current(),
```
keimlink: I would suggest to use [`django.contrib.sites.shortcuts.get_current_site() `](https://docs.djangoproject.com/en/1.10/ref/contrib/sites/#django.contrib.sites.shortcuts.get_current_site) instead.

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+            'name': name,
+            'admin_url': admin_url,
+            'description': report.description
+        })
+
+        emails.send_email_with_template(
+            recievers, 'report_moderators', context)
```
keimlink: You don't need a line break here. 😉

euth/reports/emails.py
```diff
@@ -0,0 +1,72 @@
...
+
+from euth.contrib import emails
+
+
+def send_email_to_moderators(request, report):
+
+    obj = report.content_object
+
+    if obj:
+        name = obj._meta.verbose_name.title()
```
keimlink: Are you sure you want to use `title()` here? `verbose_name` should already have proper spelling.

euth/reports/emails.py
keimlink: I agree! Added a comment to #184.

euth/reports/templates/emails/report_moderators.en.txt
```diff
@@ -5,7 +5,7 @@ We have received a report request for A ressource of type {{ name }} on {{site.n
 {% if admin_url%}
 To view and edit this resource follow this link go to the following url.
 
-{{ protocol }}://{{site.domain}}{{admin_url}}
+{{domain}}{{admin_url}}
```
keimlink: Shouldn't this use `site.domain`?

euth/comments/models.py
```diff
@@ -2,6 +2,7 @@
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
 from django.db import models
+from django.utils.translation import ugettext as _
```
keimlink: Models should always use `ugettext_lazy`, see [Django documentation: Lazy translation](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/#lazy-translations).

euth/comments/models.py
```diff
@@ -18,6 +19,9 @@ class Comment(TimeStampedModel):
     is_removed = models.BooleanField(default=False)
     is_censored = models.BooleanField(default=False)
 
+    class Meta:
+        verbose_name = _("Comment")
```
keimlink: `verbose_name_plural` is missing.

euth/contrib/emails.py
```diff
@@ -0,0 +1,31 @@
...
+                            for lang in languages])
+
+    mail = EmailMultiAlternatives(
+        subject=subject.render(context).strip(),
+        body=plaintext.render(context),
+        from_email=settings.DEFAULT_FROM_EMAIL,
+        to=receivers,
+    )
+    mail.mixed_subtype = 'related'
+    f = pkg_resources.resource_stream('euth_wagtail', 'static/images/logo.png')
```
keimlink: You should use [`from django.contrib.staticfiles.finders.find()`](https://docs.djangoproject.com/en/1.10/ref/contrib/staticfiles/#finders-module) instead.

euth/reports/models.py
```diff
@@ -0,0 +1,23 @@
...
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_pk = models.PositiveIntegerField()
+    content_object = GenericForeignKey(
+        ct_field="content_type", fk_field="object_pk")
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    description = models.TextField(max_length=1024)
+
+    class Meta:
+        unique_together = (('content_type', 'object_pk', 'user'))
```
keimlink: You should either use
```python
(('content_type', 'object_pk', 'user'),)
```
or
```python
('content_type', 'object_pk', 'user')
```
But putting brackets around the tuple does not make sense. 😉

euth/reports/models.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def __str__(self):
+        return str(self.content_type) + '_' + str(self.object_pk)
```
keimlink: I would suggest to use `format()` here to be consistent. 😃

euth_wagtail/settings/dev.py
```diff
@@ -20,3 +20,5 @@
     from .local import *
 except ImportError:
     pass
+
+PROTOCOL = 'http'
```
keimlink: I think this line has become superfluous.

tests/reports/test_report_views.py
```diff
@@ -0,0 +1,58 @@
...
+
+@pytest.mark.django_db
+def test_anonymous_user_can_not_view_reportlist(apiclient):
+    url = reverse('reports-list')
+    response = apiclient.get(url, format='json')
+    assert response.status_code == status.HTTP_403_FORBIDDEN
+
+
+@pytest.mark.django_db
+def test_authenticated_user_can_report(apiclient, user):
```
keimlink: Maybe I'm overlooking something, but IMO the test name does not match the behaviour.

euth/reports/emails.py
MagdaN: The problem here is, that as this is a model with generic FK, it could theoretically happen that there is no representation in the admin interface for this model and the reverse call would fail. But you are right I will change this.

MagdaN:
@slomo : do you want to review now or should I add the FE part first?

keimlink:
LGTM 👍

MagdaN:
if it takes to long for the emails to be sent and display the success message, we might need to add a spinner

slomo:
I would probably add rather a generic indicator for all AJAX requests from the comment / rate widget

PR #186, opened 2016-08-17
2016 08 mn add avatar upload

MagdaN:
This adds a first version of the dashboard, which also contains a profile tab (this might change as a separate profile page is planned). There users can change their avatar. 

@slomo: I added some signals to the user_management models.py. I would have put them in a separate signals.py but I somehow could not use AppConfig within the settings. Maybe you could take a look?

### DIFF
--- a/euth/dashboard/apps.py
+++ b/euth/dashboard/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class DashboardConfig(AppConfig):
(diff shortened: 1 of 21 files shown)


slomo:
Works nicely. I pulled the signals out into a separate file. So feel free to merge at your liking.

I think that our image upload widget can be improved, but this should be handled outside this pull-request.

PR #190, opened 2016-08-18
2016 08 mn refactor rates

### LINKED ISSUE #178: Should rates be anonymous
Currently our UI allows not to see who has votes for an idea or comment. The rates API exposes this information. In my opinion should either hide it in UI and API or make it visible in both:

- it looks less like an accident
- it prevents unnecessary data from being queried/processed/transmitted
- it makes it more clear to the user, that she isn't voting anonymously

### LINKED ISSUE #185: update rates
* use two different react components, one for comments one for other resources, for the comments add ratecount to comments api

MagdaN:
fixes #185 
fixes #178 

PRO
* rate-list and rate-detail api do not accept GET Requests anymore
* creator of rate is not exposed via the rate api
* still uses just one react component
* no extra requests to get the rates
* comment-API ans Rate-API now expose information on rates-count
* comment (for a authenticated user, who hast rated) looks like this:
```
{ 
    "id": 38,         
     "user_name": "admin",        
     "child_comments": [],
        "is_deleted": false,
        "rates": {
            "current_user_rate_id": 65,
            "positive_rates": 0,
            "negative_rates": 0,
            "current_user_rate_value": 0
        },
        "created": "2016-08-18T08:39:46.342312Z",
        "modified": null,
        "object_pk": 6,
        "comment": "My comment",
        "content_type": 50
}
```
  * rate detail (for a authenticated user, who hast rated) looks like this
```
{
    "id": 66,
    "meta_info": {
        "user_rate_on_same_object_value": 1,
        "negative_rates_on_same_object": 0,
        "user_rate_on_same_object_id": 66,
        "positive_rates_on_same_object": 1
    },
    "object_pk": 7,
    "value": 1,
    "content_type": 50
}
```
CONTRA:
* API might be confusing
* API not really RESTful

### DIFF
--- a/euth/comments/serializers.py
+++ b/euth/comments/serializers.py
@@ -1,6 +1,8 @@
 from django.contrib.contenttypes.models import ContentType
 from rest_framework import serializers
 
+from euth.rates import models as rate_models
+
 from .models import Comment
 
 
@@ -8,10 +10,11 @@ class CommentSerializer(serializers.ModelSerializer):
     user_name = serializers.SerializerMethodField()
     child_comments = serializers.SerializerMethodField()
     is_deleted = serializers.SerializerMethodField()
+    rates = serializers.SerializerMethodField()
 
     class Meta:
         model = Comment
-        read_only_fields = ('modified', 'created', 'id', 'user_name')
+        read_only_fields = ('modified', 'created', 'id', 'user_name', 'rates')
         exclude = ('user', 'is_censored', 'is_removed')
 
     def get_user_name(self, obj):
@@ -31,7 +34,10 @@ class CommentSerializer(serializers.ModelSerializer):
         pk = obj.pk
         children = Comment.objects.all().filter(
             content_type=content_type, object_pk=pk).order_by('created')
-        serializer = CommentSerializer(children, many=True)
+        serializer = CommentSerializer(
+            children,
+            many=True,
+            context={'request': self.context['request']})
         return serializer.data
 
     def get_is_deleted(self, obj):
@@ -39,3 +45,31 @@ class CommentSerializer(serializers.ModelSerializer):
         Returns true is one of the flags is set
         """
         return (obj.is_censored or obj.is_removed)
+
+    def get_rates(self, obj):
+        """
+        Gets positve and negative rate count as well as
+        info on the request users rate
+        """
+        user = self.context['request'].user
+        contenttype = ContentType.objects.get_for_model(obj)
+        obj_rates = rate_models.Rate.objects.filter(
+            content_type=contenttype, object_pk=obj.pk)
+        positive_rates = obj_rates.filter(value=1).count()
+        negative_rates = obj_rates.filter(value=-1).count()
+        try:
+            user_rate = obj_rates.get(user=user)
+            user_rate_value = user_rate.value
+            user_rate_id = user_rate.pk
+        except:
+            user_rate_value = None
+            user_rate_id = None
+
+        result = {
+            'positive_rates': positive_rates,
+            'negative_rates': negative_rates,
+            'current_user_rate_value': user_rate_value,
+            'current_user_rate_id': user_rate_id
+        }
+
+        return result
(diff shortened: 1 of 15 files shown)


keimlink:
If I understand correctly this functionality rates comments. If so, the wording in the code is misleading: Instead of "rate" or "rates" it should use "rating" or "ratings".

slomo:
What does value corresponds to in the second response?

MagdaN:
Its the value of the rate, its bit confusing - the way it is used, it will always be the same as user_rate_on_same_object_value, but I thought I leave it in there for completeness

slomo:
Overall I find the API design kind of strange. The part about the comments looks fine to me. The rating of ideas is rather weird. For instance: Why is that field named metadata? It seems to me to contain the actual data that I am interested in. And also the fact that I can't query the sum, over the API seems odd to me.

Anyway I voiced those concerns earlier, and would also be happy with accepting the pull-request as it is.

For the issue of bad naming (rates vs. ratings) I added a new ticket #201.

euth/rates/serializers.py
```diff
@@ -1,15 +1,42 @@
...
 
     class Meta:
         model = Rate
-        read_only_fields = ('id', 'user_name')
+        read_only_fields = ('id', 'meta_info')
         exclude = ('user', 'modified', 'created')
 
-    def get_user_name(self, obj):
-        return str(obj.user.username)
+    def get_meta_info(self, obj):
```
slomo: @keimlink This serializer contains a lot of database / model related code. Should we move this to somewhere else, or is that just fine? I could see it going either the model oder services.py?

MagdaN: I would have added it to the model, but as the request is needed, I thought I leave it in the serializer.

keimlink: I agree that it would be better to have this code in at the model. If you need any additional data, you can pass it as arguments to the method.

```python
class Rate(models.Model):
    POSITIVE = 1
    NEGATIVE = -1

    def get_meta_info(self, user):
        contenttype = ContentType.objects.get_for_model(self.content_object)
        rates = rate_models.Rate.objects.filter(
            content_type=contenttype,
            object_pk=self.content_object.pk
        )
        result['positive_rates_on_same_object'] = rates.filter(value=self.POSITIVE).count()
        result['negative_rates_on_same_object'] = rates.filter(value=self.NEGATIVE).count()

        try:
            user_rate_on_same_object = rates.get(user=user)
            result['user_rate_on_same_object_value'] = user_rate_on_same_object.value
            result['user_rate_on_same_object_id'] = user_rate_on_same_object.pk
        except:
            result['user_rate_on_same_object_value'] = None
            result['user_rate_on_same_object_id'] = None

        return result
```

Furthermore `1` and `-1` should be symbols, so that they defined at a single place. Also you should add an expression to `except` so that it only catches the exceptions you want to handle.

Please see Tom Christie's article [Django models, encapsulation and data integrity](https://www.dabapps.com/blog/django-models-and-encapsulation/) for more information about this concept.

euth/comments/serializers.py
```diff
@@ -31,11 +34,42 @@ def get_child_comments(self, obj):
...
+            user_rate_id = None
+
+        result = {
+            'positive_rates': positive_rates,
+            'negative_rates': negative_rates,
+            'current_user_rate_value': user_rate_value,
+            'current_user_rate_id': user_rate_id
+        }
+
+        return result
```
keimlink: I would suggest to change the code according to my proposal in `euth/rates/serializers.py`.

euth/rates/api.py
```diff
@@ -1,47 +1,32 @@
...
-            obj.value = 0
-            obj.save()
-            return Response(serializer.data)
-        else:
-            return Response(serializer.errors,
-                            status=status.HTTP_400_BAD_REQUEST)
+        rate.value = 0
+        rate.save()
+        serializer = self.get_serializer(rate)
+        return Response(serializer.data)
```
keimlink: According to my proposal in `euth/rates/serializers.py` I would suggest to move lines 29 and 30 to `Rate.update(value)`.

MagdaN:
@keimlink @slomo I moved the code to the model

slomo:
Okay than lets merge that guy. It is definitely better than the old functionality.

PR #203, opened 2016-08-23
rework of hero unit

vellip:
unified markup and scss of the header sections from both the organization and project page,
also removed old stuff that was left over from the former homepages hero unit


euth_wagtail/static/scss/components/_hero-unit.scss
```diff
@@ -1,37 +1,71 @@
...
-    .hero-unit-image {
-        background-repeat: no-repeat;
-        background-position: center center;
-        background-size: cover;
-    }
+    // centering the container vertically and horizontally
+    display: -webkit-box;
+    display: -ms-flexbox;
+    display: flex;
+    -webkit-box-align: center;
```
localgrr: Check our browser requirements, I think they are quite high therefor we don't need to use legacy prefixes

euth_wagtail/static/scss/components/_orgs.scss
```diff
@@ -103,15 +95,26 @@
     }
 }
 
-.org-badge {
+.org-badge-title {
     font-size: $font-size-medium;
```
localgrr: use rem mixin

euth_wagtail/static/scss/components/_orgs.scss
```diff
@@ -103,15 +95,26 @@
...
+    padding-left: 0;
+
+    li {
+        display: inline-block;
     }
 
-    &, .org-badge-image {
-        @include rem(margin-bottom, 20px);
+    a {
+        font-size: 17px;
```
localgrr: use rem mixin

localgrr:
screenshot-localhost 8000 2016-08-23 15-28-00
Great work ! Just some minor stuff - See inline comments too

PR #204, opened 2016-08-23
improved font rendering on osx

vellip:


euth_wagtail/static/scss/all.scss
```diff
@@ -2,6 +2,7 @@
 @import "scss/core/bootstrap";
 @import "scss/core/fonts";
 @import "scss/core/mixins";
+@import "scss/core/base";
```
localgrr: I'd put this in type instead, saves a new file

PR #206, opened 2016-08-23
first attemp to fix the hero unit

MagdaN:
@np @slomo: this PR https://github.com/liqd/euth_wagtail/pull/203 breaks the hero image on the homepage. I tried to fix with this PR but could not finish as I had to leave.

Maybe a better approach would be, to somehow revert #203 and update it so that the hero unit files (template, scss) for the home page do not get deleted but renamed instead.


vellip:
I will check this out, I think we can just keep the template for the home hero unit and make the scss work there too.

vellip:
@slomo, I think it's good now. The changes made are only for that particular problem of the page not working. I will adapt the new design in a new branch; I guess that's cleaner?

PR #214, opened 2016-08-24
Add social share data to organisation model

slomo:


euth/organisations/templates/euth_organisations/organisation_detail.html
```diff
@@ -6,9 +6,17 @@
...
+                  {% endif %}
+                  {% if organisation.handle_twitter %}
+                  <li><a href="https://twitter.com/{{organisation.handle_twitter}}"><i class="fa fa-twitter"></i></a></li>
+                  {% endif %}
+                  {% if organisation.handle_instagram %}
+                  <li><a href="https://www.instagram.com/{{organisation.handle_instagram}}"><i class="fa fa-instagram"></i></a></li>
+                  {% endif %}
+                  {% if organisation.webpage %}<li><a href="{{organisation.webpage}}"><i class="fa fa-link"></i></a></li>
+                  {% endif %}
                 </ul>
```
vellip: Is there a way to make sure, the <ul> won't be empty if there's none of the fields used? Probably only way to check is to repeat all the {% if %} together inside one if around the whole <ul>, mh?

slomo: Add some code fixing this and force pushed

vellip:
@slomo, think we can merge this one?

slomo:
@vellip I think you can merge this one :smile:

PR #216, opened 2016-08-25
Change permissions depending on phase

slomo:
- prevent creation of ideas by users if the phase doesn't allows creating ideas


slomo:
@keimlink I know that placing rules in the idea app, is probably not the way to go. But I'd like to have some first feedback on this pull-request.

slomo:
@vellip consider this one ready to merge

PR #230, opened 2016-08-30
Add dates to phases

slomo:


slomo:
Currently `days_left` will not return the days but a timedelta, that needs to be addressed (and tested).

PR #252, opened 2016-09-06
Move upload-wrapper class into widget definition

slomo:
Issues fixed:

- [x] `upload-wrapper` class needs to surround the form field
- [x] wrapping field in `col` class will result in buttons and labels floating arround
- [ ] ~~deleting and uploading new image doesn't work (probably anoying to fix)~~ (see #261)


slomo:
@vellip Could you have look at the second issue. I might dig into the third later on.

vellip:
Did that, and filename now gets inserted immediately too, giving instant feedback when changing images.

PR #260, opened 2016-09-07
Add dashboad ui for user requests

slomo:


vellip:
@slomo This looks good, but now there's an error when submitting the form, my changes might be causing that, though I am not sure.

PR #268, opened 2016-09-07
restyle dashboard navigation

vellip:
grey to black!


vellip:
@slomo just a heads-up: the style+markup for the organization dropdown is in this branch, not yours

PR #275, opened 2016-09-09
2016 09 mn add tests and bugfixes to documents

MagdaN:


slomo:
Should the weight and module be unique together?

PR #288, opened 2016-09-12
Invitations for private projects

slomo:


MagdaN:
I got an accept link via email, then I registered after the register process I was redirected to a the project and I was asked if I wanted to join the project. But when I clicked on 'accept' I was redirected to the project page but wan NOT a memeber.

MagdaN:
When I invite a user that has not registered, the new user receives an email and registers after registering and clicking the activation link, the user gets redirected to a page that does not exist (404)  e.g. http://localhost:8000/de/memberships/invites/621435d6-4259-45d5-adfe-8d6570046c08/accept/

MagdaN:
works now

PR #297, opened 2016-09-13
validate phase dates

vellip:


euth/phases/models.py
```diff
@@ -30,6 +31,13 @@ def __str__(self):
     def content(self):
         return content[self.type]
 
+    def clean(self):
+        if (self.end_date < self.start_date):
+            raise ValidationError({
+                'end_date': 'End date can not be smaller than the start date.'
```
slomo: If you want to you can probably also specify boths fields here. Also this string should be translated using ugettext

slomo:
It seems that a test case calling clean is missing. You probably want to test somewhere that the exception is raised at all.

PR #299, opened 2016-09-13
2016 09 mn create projects dashboard

MagdaN:
This enables to initiate projects with Idea collection or Commenting text module in the dashboard. Still missing and breaking:

- [x] at the moment editing projects in django admin breaks
- [x] the value for is_draft is always set to false
- [x] multiform should be used in update view also
- [x] when there are validation errors in one of the tabs the number of errors shall display on the tab
- [ ]  tiles for choosing Idea collection or Commenting Text Module need to be added (styling)
- [x] are there validations for the phases, that the dates are right?
- [x] datefields should have a datepicker or some helptext of how to type the date
- [ ] phase field labels are broken in create


slomo:
I moved most of the code to #323 #320 #314

PR #304, opened 2016-09-15
Permissions for commenting text

slomo:
-  includes some refactoring for ideas permissions
 - fixes user stranding on the join page


vellip:
Accessing a detail paragraph page results in a 403 response, when there are other modules active as well, causing the comment text area on the document page to be hidden as well.

slomo:
Apparently I forgot to merge a file.

PR #325, opened 2016-09-21
Selector for project blueprints

slomo:
Allows to select upon project create what phases the project has.


slomo:
This is missing proper styling but I would still merge it.

slomo:
Now it is also in conflict with #332, it should be probably merged after that one.

PR #343, opened 2016-10-06
Mn 2016 10 fix logo stylings

### LINKED ISSUE #336: organization logo styling bugs
The way orga logos are shown has some minor issues:

- [x] On the project page the logo is not shown (blank white circle): https://opin.me/de/projects/commenting-text/
- [x] The logo should have more padding from the edges of the white circle on the project page (just like it does on the tiles for an organization on the organization overview page): https://opin.me/de/orgs/conseil-parisien-de-la-jeunesse/
- [ ] Logos in bigger format on the organization page look blurry – even on non-high-res screens: https://opin.me/de/orgs/opin-examples/

MagdaN:
fixes #336 

@vellip I guess the change in _hero-unit.css has been made for a reason. If we put max-width to 60% the whole definition in _hero-unit.css can be deleted

### DIFF
--- a/euth/organisations/templates/euth_organisations/organisation_detail.html
+++ b/euth/organisations/templates/euth_organisations/organisation_detail.html
@@ -26,7 +26,7 @@
                     {% if organisation.logo %}
                     <img alt="{{ organisation.name }} logo" src="{{ organisation.logo|thumbnail_url:'organisation_thumbnail' }}">
                     {% else %}
-                    <img alt="opin logo" src="{% static "images/logo.svg" %}">
+                    <img alt="opin logo" src="{% static "images/logo.png" %}">
                     {% endif %}
                 </div>
                 <h1 class="euth-hero-unit-title">{{ organisation.name }}</h1>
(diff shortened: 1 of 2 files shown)


euth/organisations/templates/euth_organisations/organisation_detail.html
```diff
@@ -26,7 +26,7 @@
                     {% if organisation.logo %}
                     <img alt="{{ organisation.name }} logo" src="{{ organisation.logo|thumbnail_url:'organisation_thumbnail' }}">
                     {% else %}
-                    <img alt="opin logo" src="{% static "images/logo.png" %}">
+                    <img alt="opin logo" src="{% static "images/logo.svg" %}">
```
slomo: In other places we do some detection whether SVG is possible. Should we do it here as well, or is it now safe to assume that all browser support SVG?

vellip: We'd have to use background images instead.

PR #353, opened 2016-10-12
Fix permission handling

### LINKED ISSUE #342: As intiator you can see the listing of ideas but not the ideas in the listing
Probably applies to projects in draft and such not in drafts.

### LINKED ISSUE #364: Newly created projects have no moderators setup
If a project has been created it doesn't have any moderators setup. This results in the project being invalid, because moderators are an required option.

Also moderators can't be managed from the dashboard yet.

slomo:
- [x] ensure that all dashboard view redirect to login
- [x] ensure that ideas detail doesn't produce redirect loop
- [x] make idea view behaviour of private processes consistent for initiators
  - initiators can now view ideas of private processes
  - initiator of project is automatically moderator

fixes #342 
fixes #364

### DIFF
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -129,13 +129,14 @@ class ProjectUpdateForm(multiform.MultiModelForm):
 
 class ProjectCreateForm(multiform.MultiModelForm):
 
-    def __init__(self, blueprint, organisation, *args, **kwargs):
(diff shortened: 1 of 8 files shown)


MagdaN:
@slomo but this does not fix, that moderators can't be added to project in dashboard, right?

slomo:
no, but it ensures that at least one moderator is added, to keep the data consistent

PR #357, opened 2016-10-13
add timeline for project detail list

vellip:


euth/phases/models.py
```diff
@@ -17,8 +17,8 @@ def active_phases(self):
 
 
 class Phase(models.Model):
-    name = models.CharField(max_length=512)
-    description = models.TextField()
+    name = models.CharField(max_length=80)
```
slomo: Those changes require data migrations, otherwise they might fail if in the database there are already longer char fields.

slomo:
If you change fields on a model you will usually need a migration.

PR #361, opened 2016-10-18
Add basic webpack support for sass compilation

slomo:
- remove compress for base style
- add loaders to for sass, css and style
- adopt bootstrap paths


vellip:
@slomo check this out!

euth_wagtail/static/scss/all.scss
```diff
@@ -1,4 +1,4 @@
-@import "../icons/fontcustom/fontcustom.css";
+/* @import "../icons/fontcustom/fontcustom.css"; */
```
slomo: Not sure about this. Probably we might still need it.

euth_wagtail/static/scss/core/_bootstrap.scss
```diff
@@ -11,7 +11,7 @@
 // Reset and dependencies
 @import "bootstrap/normalize";
 @import "bootstrap/print";
-@import "bootstrap/glyphicons";
+// @import "bootstrap/glyphicons";
```
slomo: Neither sure what happens here.

slomo:
How does this work? You are using the extract loader to exclude the style loader from the bundle, but the extract css is not included anywhere?

euth_wagtail/static/scss/core/_bootstrap.scss
vellip: This is fine, it's actually not needed from what I've found.

slomo:
LGTM @MagdaN do you want to check this or should I merge it?

PR #372, opened 2016-10-19
add transifex

vellip:
add transifex client,
add tx config,
update translation files


slomo:
@vellip Please document how the transifex client should be used.

vellip:
@slomo Please check if everything works with the documentation now.

PR #387, opened 2016-10-21
2016 mn tests for document api

MagdaN:


euth/documents/api.py
```diff
@@ -10,5 +10,5 @@ class DocumentViewSet(viewsets.ModelViewSet):
     queryset = Document.objects.all()
     serializer_class = DocumentSerializer
     permission_classes = (
-        IsModerator,
+        permissions.IsAuthenticatedOrReadOnly, IsModerator,
```
slomo: This is fixing an exception in the moderator exception if the user isn't logged in? Can this be fixed in the `isModerator` permission?

MagdaN: Its done as in ratings and comments.

slomo: But why is it is `AuthenticatedOrReadOnly`?

MagdaN: OK, dann nehmen wir http://www.django-rest-framework.org/api-guide/permissions/#isauthenticated

PR #393, opened 2016-10-24
Mn 2016 delete comments of deleted document and paragraphs

### LINKED ISSUE #383: delete comments of paragraphs and document if document was deleted

MagdaN:
fixes #383

### DIFF
--- a/euth/documents/models.py
+++ b/euth/documents/models.py
@@ -31,6 +31,14 @@ class Document(module_models.Item):
     def paragraphs_sorted(self):
         return self.paragraphs.all().order_by('weight')
 
+    @cached_property
+    def comments(self):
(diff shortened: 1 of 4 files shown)


euth/documents/models.py
```diff
@@ -31,6 +31,14 @@ def clean(self, *args, **kwargs):
     def paragraphs_sorted(self):
         return self.paragraphs.all().order_by('weight')
 
+    @cached_property
```
slomo: Do we like to solve this in a comperable way to how it has been solved with the ideas models (e.g `GenericRelation`)?

slomo: Is this function only implemented for the tests?

MagdaN: Sure but I would put it to a seperate PR

PR #397, opened 2016-10-24
Mn 2016 10 add helptexts

MagdaN:


euth/dashboard/templates/euth_dashboard/project_form.html
```diff
@@ -31,6 +31,7 @@ <h3 class="tab-panel-title">{% trans "info" %}</h3>
                 {% if not field == form.project.result %}
                 <div class="form-group{% if field.errors %} has-error{% endif %}">
                     <label>{{ field.label }}</label>
+                    <p><i>{{ field.help_text }}</i></p>
```
vellip: The use of <i /> is [inappropriate](https://developer.mozilla.org/de/docs/Web/HTML/Element/i) here. Maybe better use a class.

MagdaN:
@vellip I added the class

Issue #381, opened 2016-10-21
add id to paragrpahs in frontend


MagdaN:

slomo:
Could elaborate here a bit more, what the issue is?

PR #398, opened 2016-10-25
improve paragraph edit form

slomo:
- add animation
- prevent creation of new paragraphs from moving
- prevent creating multiple document on first create
- lower up and down flickering by resizing textareas

solves #381 ???

### DIFF
--- a/euth/documents/static/documents/Paragraph.js
+++ b/euth/documents/static/documents/Paragraph.js
@@ -10,10 +10,10 @@ var Paragraph = React.createClass({
     this.props.deleteParagraph(this.props.index)
   },
   up: function () {
-    this.props.moveParagraphUp(this.props.index, this)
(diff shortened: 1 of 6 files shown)


MagdaN:
initial creation of document fails with: {"paragraphs":[{"weight":["Dieses Feld muss ausgefüllt werden."]}]}

MagdaN:
I think the buttons should be disabled and not invisible when moving up or down does not make sense

MagdaN:
When I changed the order of two paragrpahs and press save the order is changed back to what it was before

slomo:
@MagdaN all issues addressed, could you test again

MagdaN:
@slomo when I add a new paragraph in between two paragraphs it gets posted immediately an I get error messages saying that the headline and the text of the new paragraph should not be empty.

PR #399, opened 2016-10-25
Add generic relations to documents

### LINKED ISSUE #396: Use Generic Relation everywhere where generic fks are used

slomo:
~~fixes #396~~

### DIFF
--- a/euth/comments/static/comments/CommentForm.jsx
+++ b/euth/comments/static/comments/CommentForm.jsx
@@ -35,11 +35,16 @@ let CommentForm = React.createClass({
           <input type="submit" value={django.gettext('post')} className="submit-button" />
         </form>
       )
-    }
-    if (!this.context.isAuthenticated) {
+    } else if (!this.context.isAuthenticated) {
       return (
         <div className="comments_login">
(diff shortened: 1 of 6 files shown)


MagdaN:
Should we not use a generic relation in comment to get the child comments?

vellip:
I can't add a paragraph filling out the (actually optional) headlines.

vellip:
When the create phase is over and the paragraphs should be shown, there is an error: "'RelatedManager' object is not iterable" at "{% for paragraph in object.paragraphs %}"

vellip:
Being a normal user, I don't get the comment box on the whole document_detail, only for the detail paragraph view. (There's probably some component returning nothing/undefined in jsx)

vellip:
Also, the comment count is not shown (renders 0), needs an `.all` in document_detail:17

slomo:
@vellip you can also leave line comments ;)

slomo:
@vellip @MagdaN addressed your issues

slomo:
@MagdaN generic relations for comments will be on a different pull request

MagdaN:
@slomo are you going to add an new issue?

PR #400, opened 2016-10-25
Use json in all rest APIs

### LINKED ISSUE #374: fix function in api.js to use contenttype json and fix all react components

slomo:
Fix #374


MagdaN:
works

PR #406, opened 2016-10-25
add verbose_name to flashpoll model

### LINKED ISSUE #370: Specify "Key" in creating mobile polling project
Hey A-Team, setting up a mobile polling works, thanks for that. For a better understanding how to import the poll from Flashpoll, namely via the ID, we might want to add that information to the create project field. Maybe change "key" to "Flashpoll ID" and add a screenshot of it where to find it? Thanks a lot :) 

bildschirmfoto 2016-10-19 um 13 20 09

MagdaN:
fixes #370


slomo:
Hi, was hälst du davon die Migration umzubennen? In irgendetwas wie `add_verbose_name_to_token` oder so? Dann kann man beim Blick in en Migrationordner einfacher sehen, was welche Migration macht.

MagdaN:
Yes, we can try that, but this will only work if we change just one thing, if there are several things it might get hard to find a good filename and I would not want to create one separate migratinofile for each change.

MagdaN:
@slomo I changed the filename, can you merge?

PR #408, opened 2016-10-26
avoid encoding get and head parameters as json

slomo:


MagdaN:
It works, but we could have used jquerys GET-Shorthand Method instead, maybe more robust?

PR #412, opened 2016-10-26
Unify user attributes for various types

### LINKED ISSUE #184: FK-Fields to a user in Models should all have the same name
We different models which all save a user as creator, but the fields are named differently (user, creator, ...) 
I would suggest to always call them creator. Maybe we could also implement a abstract model with a creator field and make the other models inherit from this abstract model.

slomo:
fixes #184

### DIFF
--- a/euth/comments/api.py
+++ b/euth/comments/api.py
@@ -1,7 +1,7 @@
 from rest_framework import filters, permissions, viewsets
 from rest_framework.response import Response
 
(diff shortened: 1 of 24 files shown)


MagdaN:
@slomo can you please merge?

slomo:
@MagdaN merged

MagdaN:
LGTM

PR #417, opened 2016-10-27
Mn 2016 10 style project form

### LINKED ISSUE #379: create project in dashboars: info texts and headlines
The form labels and descriptions in the info tab are very important as there is info for the project in general and the info tab that needs to be given. I created some new labels, descriptions and placeholder texts as well as two headlines in the forms that will hopefully explain it a bit better to users.

MagdaN:
fixes #379


slomo:
@MagdaN I changed to smaller things feel free to merge

PR #420, opened 2016-10-28
Mn 2016 10 display info for project that have not started

MagdaN:
still missing the image, but I would suggest to add the image in a later PR.


slomo:
Works fine, but I have two questions:

1. Why is the include in contrib and not in the projects app?
2. Why is the empty phase name and empty phase description display?

MagdaN:
1. @vellip removed the styles from projects to a separate file (components/_info_note.scss), so that it can be used in other places, too. So it felt wrong to leave the html file in projects.

MagdaN:
2. sure, we could add an if clause around view.phase.name and view.phase.description if that suits you better

PR #432, opened 2016-11-02
Refactor last phase handling

slomo:
- ensure that after the last phase of project the text doesn't ask to wait for the next phase
- move more behaviour from the project mixin to project model (fat model)
- fix project factory


MagdaN:
LGTM

PR #438, opened 2016-11-02
improve timeline behavior

vellip:
properly hide popover after dismissing them by clicking somewhere else,
make the dots bigger on mobile


MagdaN:
Works nicely

PR #445, opened 2016-11-03
Mn 2016 11 add pagination

### LINKED ISSUE #425: Display sorting dropdown in idea list only if there are more than one entries

MagdaN:
Fixes #425 
Adds pagination to idea list.


euth/ideas/templates/euth_ideas/idea_list.html
```diff
@@ -37,5 +37,53 @@
     {% include "euth_ideas/includes/idea_list_tile.html" with idea=idea %}
     {% endfor %}
 
+    {% if is_paginated %}
+    <nav aria-label="Page navigation">
```
MagdaN: Its a nav because it is a nav in Twitter Bootstrap: http://getbootstrap.com/components/#pagination

PR #428, opened 2016-11-01
add (s)css-lint, smaller refactoring, add lint step to npm

vellip:
renamed classes to suit unified naming style,
move several styles into own files,
new directory for forms in components,
remove redundant id from slider

PR #447, opened 2016-11-03
css linting and small refactoring

slomo:
- add (s)css-lint
 - add lint step to npm
 - renamed classes to suit unified naming style
 - move several styles into own files
 - new directory for forms in components
 - remove redundant id from slider

This is #428 rebased on the current master


slomo:
There is currently a bug with the hero unit on the organisation page.-

PR #448, opened 2016-11-04
replace usage of organisation name with title

slomo:
- allows user to set display name of organisation
- allows user to translate display name of organisation


MagdaN:
LGTM

PR #452, opened 2016-11-07
add doc for css guidelines

vellip:


slomo:
Very nice. Reading about it, it seems that we might consider to somehow differentiate between modifier and children. Do you think this can be useful, or rather to complicated?

vellip:
I think it would be useful! I am however not sure, if we find a good way
that everyone is happy with. Bootstrap does make no difference and if we
deviate from Bootstrap we could also just use BEM again.

A simpler/more bootstrap-close solution might be to add an "is" in front of
the modifier, like `.tweet-title-islarge` I'd rather not add a dash after
the "is" though, because it feels somewhat inconsistent that way.

Otherwise we might just look at BEM for this one and use
`.tweet-title--large`.

I don't mind either, just wanting to say, this is not what Bootstrap does,
so the argument to use Bootstrap as reference gets somewhat invalid then.
Though it would still be closer to Bootstrap than BEM.

Let's talk about it tomorrow.

vellip:
After discussing it: We won't make a difference between modifier and child element.

MagdaN:
I think this very good. Thanks, for creating it. Should I merge?

vellip:
I am fine with it!

PR #454, opened 2016-11-08
add stylelint-order and ruleset for grouped properties

vellip:


MagdaN:
@vellip @slomo should we fix the warnings in this branch?

MagdaN:
Well I started to clean some files

vellip:
Shall we discard this idea then? It won't easily work with the `@include rem(...)` mixin.

PR #461, opened 2016-11-10
moderator label and label/badge refactoring

vellip:
show moderator label in comments and ideas,
rename badge classes to labels for semantics,
add another label,
fix indentation


euth/comments/static/comments/Comment.jsx
```diff
@@ -99,14 +99,19 @@ var Comment = React.createClass({
   },
 
   render: function () {
-    var CommentList = require('./CommentList')
+    let CommentList = require('./CommentList')
```
slomo: shouldn't this be `const`?

PR #470, opened 2016-11-11
Mn 2016 11 add actions

MagdaN:


tests/reports/test_report_views.py
```diff
@@ -50,6 +50,7 @@ def test_authenticated_user_can_post(apiclient, user, comment, admin):
     response = apiclient.post(url, data, format='json')
     assert response.status_code == status.HTTP_201_CREATED
     assert response.data['description'] == 'This comment sucks'
-    assert len(mail.outbox) == 2
-    assert 'A Comment has been reported' in mail.outbox[0].subject
-    assert 'A Comment that you created' in mail.outbox[1].subject
+    assert len(mail.outbox) == 3
+    assert 'A Comment was added to your' in mail.outbox[0].subject
+    assert 'A Comment has been reported' in mail.outbox[1].subject
+    assert 'A Comment that you created' in mail.outbox[2].subject
```
slomo: Why is this creating an notification?

euth/comments/signals.py
```diff
@@ -1,12 +1,43 @@
...
+from euth.actions.models import Action
 from euth.contrib import services
 
+from euth.projects.models import Project
+
 from .models import Comment
 
 
+@receiver(post_save, sender=Comment)
+def add_action(sender, instance, created, **kwargs):
```
slomo: Shouldn't this code be in the actions app? Otherwise we are required to change all other apps in order to have actions tracked.

euth/actions/emails.py
```diff
@@ -0,0 +1,20 @@
+from euth.contrib import emails
+
+
+def notify_creator_on_create_action(action):
```
slomo: Couldn't this be based somehow on `euth.contrib.email.UserNotification?

tests/reports/test_report_views.py
MagdaN: Because here a comment is added and this comment gets reported

slomo: But where does it add an comment here? In the factory? If so I would propose to mute signals for the factory. http://factoryboy.readthedocs.io/en/latest/orms.html#disabling-signals

MagdaN:
@slomo I muted the signals of the comments factory

euth/actions/emails.py
slomo: The thing here is, I am trying to deprecate the Email sending via the `send_email_with_template` function and replace it with the class-based emails (comparable to class-based views). Currently there are two places left: `reports`  and `users`. Reports can be changes as soon as the report dashboard tab is ready.

MagdaN: OK, what should happen now?

PR #471, opened 2016-11-11
fix slider for safari

### LINKED ISSUE #441: carousel on home page extends beyond page width
Currently the carousel on https://opin-dev.liqd.net/de/ extends beyond the page with and creates an unsightly overhang that scrolls. The problem seems to be only present in Safari (tested on version 10). In Firefox 47 the carousel seemed fine.
bildschirmfoto 2016-11-03 um 10 11 16

vellip:
let the slider start at 0
fixes #441


MagdaN:
fixes the issue

PR #473, opened 2016-11-11
validate aspect ratio

slomo:


MagdaN:
@slomo the helptexts should be changed, too. Right?

MagdaN:
works

slomo:
@MagdaN updated the helptexts

PR #477, opened 2016-11-15
Mn 2016 11 notify followers

MagdaN:
to test this: 

- add followers to project

- set enddate of one phase of that project to end within the next 24 h

- run `python manage.py notifyfollowers`


vellip:
Looking good!

PR #478, opened 2016-11-16
Mn 2016 11 remove comment list

### LINKED ISSUE #462: Comments/documents of private projects are available using the API
Private content is not properly filtered, resulting in the following issue:

 - unauthenticated users can request all comments using the API
 - authenticated users can request all documents using the API

MagdaN:
fixes #462

### DIFF
--- a/euth/comments/templatetags/react_comments.py
+++ b/euth/comments/templatetags/react_comments.py
@@ -4,7 +4,9 @@ from django import template, utils
 from django.contrib.contenttypes.models import ContentType
 from django.utils.safestring import mark_safe
 
+
 from ..models import Comment
+from ..serializers import ThreadSerializer
 
 register = template.Library()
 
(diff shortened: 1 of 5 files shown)


euth/comments/templatetags/react_comments.py
```diff
@@ -13,6 +15,10 @@
 def react_comments(context, obj):
     request = context['request']
 
+    serializer = ThreadSerializer(
+        obj.comments.all(), many=True, context={'request': request})
```
MagdaN: @slomo can we assume that everybody who wants to use the comments adds a comments generic relation to their model?

slomo: I think that wouldn't be to wired to request. If it turns out to be a problem at some point, than we can add an if here.

euth/documents/api.py
```diff
@@ -1,12 +1,16 @@
...
 
 from .models import Document
 from .serializers import DocumentSerializer
 
 
-class DocumentViewSet(viewsets.ModelViewSet):
+class DocumentViewSet(mixins.CreateModelMixin,
+                      mixins.RetrieveModelMixin,
+                      mixins.UpdateModelMixin,
+                      mixins.DestroyModelMixin,
```
MagdaN: Yes, you are right. It should be ok to remove them.

MagdaN:
@slomo I removed retrieve and destroy

PR #481, opened 2016-11-16
add manual following for projects

slomo:


MagdaN:
@slomo this is not working at, did you maybe forget to commit something?

PR #482, opened 2016-11-16
Update python dependencies

slomo:
- drop all unused deps
- update wagtail to 1.7
- update some other django related packages


MagdaN:
works, but I needed to run 'migrate'

PR #483, opened 2016-11-16
Mn 2016 11 add error page

MagdaN:
to test this: 

- set DEBUG=False in settings/dev.py
- start server with: python manage.py runserver --insecure


euth_wagtail/templates/500.html
```diff
@@ -1,17 +1,6 @@
...
-<!--[if gt IE 8]><!--> <html class="no-js"> <!--<![endif]-->
-    <head>
-        <meta charset="utf-8" />
-        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-        <title>Internal server error</title>
-        <meta name="viewport" content="width=device-width, initial-scale=1" />
-    </head>
-    <body>
-        <h1>Internal server error</h1>
+{% extends "base_errors.html" %}
```
MagdaN: @slomo this won't work, right?

slomo:
I think your first approach to not extend `base.html` for the `500.html` was the right approach. Sorry for the inconvenience.

slomo:
I added a reduced 500 page, feel free to merge

PR #488, opened 2016-11-18
my project view in dashboard

slomo:


vellip:
After talking about it, we decided to move that list to the user profile instead, so this one right here is obsolete now.

PR #505, opened 2016-11-23
Fix lint errors

xi:
I got the following error message:

    .../euth_wagtail/euth/documents/static/documents/ParagraphBox.jsx
      113:11  error  Closing curly brace should be on the same line as opening curly brace or on the line after the previous block  brace-style
      140:11  error  Closing curly brace should be on the same line as opening curly brace or on the line after the previous block  brace-style


vellip:
Thanks

Issue #495, opened 2016-11-21
All recipients visible in email when project is deleted


MagdaN:
email adds all recipients to 'to'

vellip:
Mails are now sent one by one.

MagdaN:
issue is not fixed

slomo:
I added a pr for this to adhocracy4 (https://github.com/liqd/adhocracy4/pull/36)

slomo:
PR has been merged need to update A4 version

slomo:
Solved with #695

PR #508, opened 2016-11-23
Mn 2016 11 notify moderators on idea added

MagdaN:
Within this PR I added a field _recipients_ to the action model. A list of email addresses (of the recipients ) is saved as JSON in this field. By this we can move logic (e.g. get recipients and check if the get_notification flag is set  or if the creator of a comment is not the creator of the idea etc.) from the Action to the models and apps, which create the Action instance. 

I am also against moving the creation of Actions from the signals in comments and ideas to Actions to keep the Action App slim and dont add too many dependencies to other apps.

I am just not sure, if we should save a JSON in the Database?

### DIFF
--- a/euth/actions/emails.py
+++ b/euth/actions/emails.py
@@ -1,3 +1,5 @@
+import json
+
 from django.contrib import auth
 from django.utils import translation
 
@@ -19,8 +21,10 @@ def notify_creator_on_create_action(action):
         'url': url,
     }
 
+    jsonDec = json.decoder.JSONDecoder()
+
     emails.send_email_with_template(
-        [action.target.creator.email], 'notify_creator', context)
+        jsonDec.decode(action.recipients), 'notify_creator', context)
 
 
 def notify_followers_on_almost_finished(project):
--- a/euth/actions/models.py
+++ b/euth/actions/models.py
@@ -43,6 +43,8 @@ class Action(models.Model):
     verb = models.CharField(max_length=255, db_index=True)
     description = models.TextField(blank=True, null=True)
 
+    recipients = models.TextField(blank=True, null=True)
+
     def __str__(self):
 
         ctx = {
--- a/euth/actions/migrations/0003_action_recipients.py
+++ b/euth/actions/migrations/0003_action_recipients.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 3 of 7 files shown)


euth/actions/emails.py
```diff
@@ -19,8 +21,10 @@ def notify_creator_on_create_action(action):
         'url': url,
     }
 
+    jsonDec = json.decoder.JSONDecoder()
+
     emails.send_email_with_template(
-        [action.target.creator.email], 'notify_creator', context)
+        jsonDec.decode(action.recipients), 'notify_creator', context)
```
slomo: This will sent one email with all recipients in the `To:`-header, I think we should not do this.

euth/actions/models.py
```diff
@@ -43,6 +43,8 @@ class Action(models.Model):
     verb = models.CharField(max_length=255, db_index=True)
     description = models.TextField(blank=True, null=True)
 
+    recipients = models.TextField(blank=True, null=True)
```
slomo: Why do we want to store the recipients of an action? Shouldn't the recipients be the same for each action?

slomo:
Regarding the decoupling of other apps and actions: In this approach we might not depend on each app from the action app, but now we depend on the action app from each app. Also this spreads out the code for implementing notifications/actions quite a lot. In addition the signal code is quite repetitive and will be required for every new model.

What I proposed instead, and maybe I didn't made that clear: 

We move the signals setup to the action app, by either hard coding a list of models that should be watched or reading that list from config.

``` actions/signals.py
actionable_models = [
    ('euth_ideas', 'Idea'),
    ('euth_documents', 'paragraphs')
]

def add_action(sender, instance, created, **kwargs):
   pass  # implement generic action creation here

for app, model in actionable_models:
   post_save.connect(add_action, apps.get_model(app, model))
```

euth/actions/emails.py
MagdaN: Yeah right, I also added an issue here: #495 (should that be changed in the send email with template function?)

euth/actions/models.py
MagdaN: I wanted to remove the logic of getting the recipients from the actions app. The recipients are not always the same, e.g. for 'create': for comments it is the creator of the resource that has been commented but for idea it is the moderators of the project.

MagdaN:
We can implement it the way you suggest, but how do we get the list of recipients?

PR #512, opened 2016-11-24
Mn 2016 example resource

MagdaN:
This is the sample PR that @xi and @2e2a requested. It shows how to add a new Item with phases and how to display it in the dashboard as blueprint. The following is happening (be aware that except for 5 and 6 everything is pure Django):

1) add new app (optional, you could also add your code to an existing app)
2) add model ( you need to run 'python manage.py makemigrations proposals' to create the migrations, run 'python manage.py migrate' afterwards)
3) add admin (now you can create Proposals via /admin)
4) add urls
5) add phases (the projectmixin adds the project context to the view)
6) add Blueprint
7) add Proposal Detail

@xi and @2e2a : please say what we should add next!


slomo:
@xi @2e2a @MagdaN I would close the pull-request, but not delete the branch, so feel free to continue your work here.

PR #513, opened 2016-11-24
Pv 2016 11 user profile

vellip:
is based on #511, merge that first for correct change view

### DIFF
--- a/euth/users/views.py
+++ b/euth/users/views.py
@@ -1,3 +1,5 @@
+from django.db import models as django_models
+from django.utils import timezone
 from django.views.generic.detail import DetailView
 from euth.projects.models import Project
 from . import models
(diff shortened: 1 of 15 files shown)


MagdaN:
Looks cool, only if I dont follow anyting, it looks like that:

bildschirmfoto 2016-11-24 um 17 10 28

euth/users/views.py
```diff
@@ -9,4 +11,15 @@ class ProfileView(DetailView):
...
     @property
     def get_participated_projects(self):
-        return Project.objects.filter(participants=self.object)
+        user = self.object
+
+        qs = Project.objects.filter(
+            django_models.Q(
+                module__phase__end_date__gt=timezone.now()
+            ),  # only active projects
+            django_models.Q(follow__creator=user) |
```
MagdaN: You have to filter for enabled=True also, otherwise the project will not be removed from the list when you unfollow and reload.

euth/users/views.py
```diff
@@ -9,4 +11,15 @@ class ProfileView(DetailView):
 
     @property
     def get_participated_projects(self):
-        return Project.objects.filter(participants=self.object)
+        user = self.object
+
+        qs = Project.objects.filter(
+            django_models.Q(
+                module__phase__end_date__gt=timezone.now()
```
MagdaN: Should also finished project been shown? Designer says yes

vellip:
Moritz said no.

MagdaN:
@damusp @rittermo what should we do? At the moment I can follow finished projects - so it feels weird that they don't appear in my list.

damusp:
I really think they should appear in the list. 

As a user I would like to see the projects that I have been participating in the past, even if I don't interact with them as much anymore. Wouldn't it be sad if they just vanished? Think about all the good memories!

PR #523, opened 2016-11-29
Pv 2016 11 change follow button and update profile

vellip:
add follower count to button,
add btngroup,
add template tag to follows to check if following,
add following and member indication to profile's project list,
update menu

### DIFF
--- a/euth/follows/serializers.py
+++ b/euth/follows/serializers.py
@@ -1,3 +1,4 @@
+from django.db.models import Q
 from rest_framework import serializers
 
 from euth.projects import models as prj_models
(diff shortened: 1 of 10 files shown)


euth/follows/serializers.py
```diff
@@ -8,6 +9,7 @@
 class FollowSerializer(serializers.ModelSerializer):
 
     project = serializers.SlugRelatedField(read_only=True, slug_field='slug')
+    follows = serializers.SerializerMethodField()
```
slomo: More informal than a change request: There is a similar approach in the [rating api](https://github.com/liqd/euth_wagtail/blob/master/euth/ratings/serializers.py#L7), where also additional information is serialised as `meta_info` dict.

euth/follows/serializers.py
```diff
@@ -23,3 +25,9 @@ def save(self, *args, **kwargs):
             project = prj_models.Project.objects.get(slug=slug)
             kwargs['project'] = project
         return super().save(*args, **kwargs)
+
+    def get_follows(self, obj):
+        return models.Follow.objects.filter(
+            Q(project=obj.project),
```
vellip: Done!

PR #526, opened 2016-11-30
add penguin avatars

vellip:


slomo:
I think we should try to keep the avatar fallback out of most of the templates, by either introducing a avatar_url property or a template tag.

Also we probably want to make sure that the user avatar always has a white background.

vellip:
I made a template tag.

PR #539, opened 2016-12-02
Add auto follow

slomo:


euth/follows/signals.py
```diff
@@ -0,0 +1,17 @@
...
+        models.Follow.objects.get_or_create(
+            project=instance.project,
+            creator=instance.creator,
+            defaults={
+                'enabled': True,
+            })
+
+
+post_save.connect(autofollow_hook, 'euth_comments.Comment')
+post_save.connect(autofollow_hook, 'euth_ideas.Idea')
```
MagdaN: How is this going to work if follows is in core? Then the connectors also need to be added to the apps in the custom platforms, right?

PR #548, opened 2016-12-07
Mn 2016 11 notify moderators on idea added2

MagdaN:
Notify moderators without extra field in action table

### DIFF
--- a/euth/actions/signals.py
+++ b/euth/actions/signals.py
@@ -35,6 +35,20 @@ def add_action(sender, instance, created, **kwargs):
         action.target_object_id = instance.object_pk
         action.project = project
         action.save()
+
+    elif isinstance(instance, Item):
+        project = instance.module.project
+        project_contenttype = ContentType.objects.get_for_model(project)
+
+        action.target_content_type = project_contenttype
+        action.target_object_id = project.pk
+        action.project = project
+        action.save()
+
+    else:
+        pass
+
+
 for app, model in settings.ACTIONABLE:
     post_save.connect(add_action, apps.get_model(app, model))
     if instance.verb == verbs.COMPLETE:
--- a/euth/actions/signals.py
+++ b/euth/actions/signals.py
@@ -1,19 +1,31 @@
+from django.apps import apps
+from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
 from euth.actions import emails
+from euth.modules.models import Item
+from euth.projects.models import Project
 
 from . import verbs
 from .models import Action
 
 
-@receiver(post_save, sender=Action)
-def send_notification(sender, instance, created, **kwargs):
-
-    if instance.verb == verbs.CREATE and hasattr(instance.target, 'creator'):
+def notify_creator(instance):
+    if hasattr(instance.target, 'creator'):
         creator = instance.target.creator
         if creator.get_notifications and not creator == instance.actor:
-            emails.notify_creator_on_create_action(instance)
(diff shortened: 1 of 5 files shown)


euth/actions/signals.py
```diff
@@ -1,18 +1,75 @@
...
+    actor = instance.creator
+
+    action = Action(
+        actor=actor,
+        verb=verb,
+        action_object_content_type=content_type,
+        action_object_object_id=instance.pk,
+    )
+
+    if str(content_type) == 'Comment':
```
slomo: Comments do also have a project property. So they probably don't need any special treatment. when getting the project:

```
if hasattr(instance, 'project'):
    project = instance.project
else:
   project = None
```

Should be sufficient for all Actionables

euth/actions/signals.py
```diff
@@ -1,18 +1,75 @@
...
+
+def add_action(sender, instance, created, **kwargs):
+    content_type = ContentType.objects.get_for_model(instance)
+    verb = verbs.CREATE if created else verbs.UPDATE
+    actor = instance.creator
+
+    action = Action(
+        actor=actor,
+        verb=verb,
+        action_object_content_type=content_type,
```
slomo: `action_object =  instance` should to the same here

euth/actions/signals.py
```diff
@@ -1,18 +1,75 @@
...
+    elif isinstance(instance, Item):
+        project = instance.module.project
+        project_contenttype = ContentType.objects.get_for_model(project)
+
+        action.target_content_type = project_contenttype
+        action.target_object_id = project.pk
+        action.project = project
+        action.save()
+
+    else:
```
slomo: Empty else clause can probably be left out.

euth/actions/signals.py
```diff
@@ -1,18 +1,75 @@
...
+def notify_creator(instance):
+    if hasattr(instance.target, 'creator'):
+        creator = instance.target.creator
+        if creator.get_notifications and not creator == instance.actor:
+            emails.notify_users_on_create_action(instance, [creator.email])
+
+
+def notify_moderators(instance):
+    if str(instance.target_content_type) == 'project':
+        recipients = instance.target.moderators.all().values_list(
```
slomo: Should the actor be filtered out from this list?

euth/actions/signals.py
```diff
@@ -1,18 +1,75 @@
...
 
+def notify_creator(instance):
+    if hasattr(instance.target, 'creator'):
+        creator = instance.target.creator
+        if creator.get_notifications and not creator == instance.actor:
+            emails.notify_users_on_create_action(instance, [creator.email])
+
+
+def notify_moderators(instance):
+    if str(instance.target_content_type) == 'project':
```
slomo: In my personal opinion `instance.target_content_type.model_class() is Project` would be preferable, but that is certainly only a question of style.

euth/actions/verbs.py
```diff
@@ -1,7 +1,7 @@
 # http://activitystrea.ms/registry/verbs/
 
-CREATE = 'create'
-UPDATE = 'update'
+CREATE = 'created'
```
MagdaN: I thought it should read like a sentence, but I don't mind.

euth/actions/signals.py
MagdaN: Go for it

euth/actions/signals.py
MagdaN: true

euth/actions/signals.py
MagdaN: ok

MagdaN:
@slomo feel free to change

PR #554, opened 2016-12-13
Mn 2016 define map for module

MagdaN:
This PR implements:

- fix settings for module so that modules can have different settings (a property in modules is added that gets the right settings)
- fix error number display in project form template so that also the number of errors in the settings module is displayed (I added a template tag that adds two numbers)

- Implements geojson fields with widgets

- Implements an Idea Collection that uses the new geojson fields (MapIdea inherits Idea, if we stick to this approach, we could add some refactoring, e.g. that both Models use the same template)


slomo:
Works for me, so far. Only some small code cleanness issues.

euth/maps/static/js/custom/map_choose_point.js
```diff
@@ -0,0 +1,96 @@
...
+    'weight': 2,
+    'opacity': 1,
+    'fillOpacity': 0.2
+  }
+
+  var basePolygon = L.geoJson(polygon, {style: polygonStyle}).addTo(map)
+  map.fitBounds(basePolygon)
+
+  var marker
+  var latlng
```
MagdaN: linter

PR #562, opened 2016-12-20
Pv 2016 12 add manuals

vellip:
1. Create a ManualIndex page under the HomePage,
2. Add a ManualsSection page under the ManualsIndex (color and image are used to show the tiles on the ManualIndex). **OR** Add a ManualsDetail page under the ManualsIndex (color and image are used to show the tiles on the ManualIndex)
3. Add a ManualsDetail page under the ManualsSection page. 
4. Create a PageCollection snippet with the previously added ManualDetail pages.
5. Add the Snippet in the ManualsSection body part.


home/models.py
```diff
@@ -136,6 +209,16 @@ class Meta:
...
+class PageCollectionBlock(core_blocks.StructBlock):
+    page_collection = snippet_blocks.SnippetChooserBlock(
+        required=False, target_model=PageCollection)
+
+    class Meta:
+        template = 'home/blocks/page_collection_block.html'
+        icon = 'snippet'
+        label = 'Page Collection'
+
+
```
MagdaN: You could add a list block with PageCollectionBlock to be displayed on the section page

home/models.py
```diff
@@ -584,3 +668,286 @@ class SimplePage(Page):
...
+    body_sv = StreamField(block_types, null=True,
+                          blank=True, verbose_name="body")
+    body_sl = StreamField(block_types, null=True,
+                          blank=True, verbose_name="body")
+    body_da = StreamField(block_types, null=True,
+                          blank=True, verbose_name="body")
+
+    subpage_types = [
+        'home.ManualsSectionPage',
+        'home.ManualsDetailPage'
```
MagdaN: Why can I add a ManualsDetailPage here? I tried it, but the styling seems to be broken. Maybe instead we could add more available blocks to the Section Page, then users could add text also

PR #563, opened 2016-12-20
flashpoll management integration initial commit

ewelle:
Modifications have been mainly done on the dashboard form and view, as well as other settings.

Note: The integration is not complete yet and still need some polishing, but the survey's creation and
publication work fine.


slomo:
@ewelle Please rebase on current master or merge with current master. This pull-request undoes a lot of the stuff that we have done on the previous weeks. Also travis is a bit unhappy, but that might be related.

slomo:
Also I think you should probably not commit, highcharts, highmaps and highstock into our statics folders. They should be installed via npm (by adding them to `package.json`). Afterwards they can be either loaded by adding them to the webpack vendor bundle (see `webpack.config.js`) or by adding them to the django static file directories (required options can be found in `euth_wagtail/settings/base.py`).

PR #593, opened 2017-01-16
Pv 2017 01 user search field

vellip:
Adds the api and such, plus styling for the typeahead.

-----
Do npm install.
To use the widget change a normal TextInput to the UserSearchInput.

For example: https://github.com/liqd/euth_wagtail/commit/d910f649d0f8315add0b9c8e2c5f734834d5ce71#diff-a970535f4d898bc5ead14851d8d0a522

I removed it afterwards because it doesn't belong there, was just for testing. This is only the widget on it's own, it's not doing anything yet. That's coming later.


MagdaN:
This works nicely, just one thing: please make sure that only users that are logged in have access to the user api

vellip:
You're right! Done.

PR #605, opened 2017-01-20
allow invitation acceptance only for same emails,

### LINKED ISSUE #592: Wrong user in project
When I get an invitation to a project via email and I am logged in with a different user and click one the link in the email, the second user is added to the project

vellip:
update test for invitation, fixes #592


slomo:
As I stated in https://github.com/liqd/euth_wagtail/issues/592#issuecomment-280573949 I think this pull-request is the total antifeature, I would like to revert it.

PR #649, opened 2017-02-06
remove baloo bhai, use open sans only

vellip:


slomo:
How is OpenSans loaded? Shouldn't it be added to the package.json or something similar?

vellip:
No, it's already in the project. We only use one font now.

PR #658, opened 2017-02-07
Dashboard organisation form allow deletion

### LINKED ISSUE #300: Organisation form doesn't allow to delete languages
This feature requires parler to release a new version with the <model>.delete_translation method.

slomo:
Fixes #300

### DIFF
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -1,6 +1,7 @@
 import collections
 import re
 
+import parler
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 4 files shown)


MagdaN:
Works nicely, except for a react error: Warning: LanguageSwitch is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

Also, maybe it would be easier to understand to add a minus-button instead of clicking the plus-button instead?

slomo:
As far as I understand mentioned react error is immanent to the implementation decision taken here in the first place. As far as I understand there is some jquery manipulation going on from inside react to stuff outside the react controlled array. I think @vellip is the expert here.

I think I agree with the minus button. But it was not in the scope of this ticket and has passed review before. If needed we should craft a small story out of it. I think we should check back with @rittermo @NadineKarbach or

vellip:
Fixed the warning. The jQuery is/wasn't the cause of the message though. Rather the fact that the label evoked a change of state without React knowing.

MagdaN:
I will create an issue for the minus button

PR #660, opened 2017-02-07
add email notification to newly added moderator

vellip:


MagdaN:
works nicely, we should consider using celery at some point.

PR #664, opened 2017-02-08
Adopt styling to new rating widget markup

slomo:


slomo:
Other update A4 adoptions should happen first.

PR #687, opened 2017-02-17
fix adding moderators

### LINKED ISSUE #680: Assign moderator role: Adding moderators does not work
bildschirmfoto 2017-02-16 um 17 24 10
When I want to add a moderator to the project (here, for example https://opin-dev.liqd.net/de/dashboard/dolorum-recusandae/projects/ducimus-labore-unde/users), I click on the search result, but the user is not added. Tested in Safari and Firefox, both don’t work.

vellip:
fixes #680 

Bug was introduced because every entry point in webpack now is named adhocracy4 and the exports of each entry point doesn't get merged with the previous ones but overridden.
In this case there was an additional entry point which overrode the opin js and evoked an error.
Long term wise we should address this differently, for example the vendor entry point exports its content into an adhocracy4 variable still, we are only lucky, that the vendors get loaded before our js gets loaded and the variable is reassigned then.

### DIFF
--- a/euth/dashboard/templates/euth_dashboard/project_users.html
+++ b/euth/dashboard/templates/euth_dashboard/project_users.html
@@ -1,5 +1,5 @@
 {% extends "euth_dashboard/base_organisation.html" %}
-{% load widget_tweaks i18n react_user_list avatar %}
+{% load widget_tweaks i18n react_user_list avatar static %}
 
 {% block extra_js %}
     {{ block.super }}
--- a/euth/users/static/users/js/user_search.js
+++ b/euth/users/static/users/js/user_search.js
@@ -1,7 +1,6 @@
-var $ = require('jquery')
-
-class UserSearch {
-  constructor (typeaheadElem) {
+/* global $ */
+(function () {
+  var UserSearch = function (typeaheadElem) {
     this.element = typeaheadElem
     this.$element = $(typeaheadElem)
     this.identifier = this.$element.data('identifier')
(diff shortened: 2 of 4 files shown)


slomo:
What do you think of the following fix: 397a4e62a8ad7efeee89364b67cfbebc8e1954a5 ?

vellip:
I guess it would work too but you would end up with a window.opin object and a window.adhocracy4 object which both have the same contents. It also seems a bit like a hack, doesn't it?

MagdaN:
Can this be merged? It tested it and fixes the issue.

slomo:
I wouldn't merge this because it fixes the wrong problem in my opinion. But if @vellip has another opinion here I would accept it.

MagdaN:
OK, @vellip @slomo could you please come up with a decision

vellip:
Well, the easiest solution would be to rename the entry point `opin` to `adhocracy4` and move back to naming the entry point's file like the name of chunk with `[name].js`.

I can understand that this is not entirely desirable. The other option I see is to split the `opin` entry point so you end up with one that is called `opin` and contains only opin relevant modules and move the adhocracy4 files into another entry point for themselves. 
Then you'd still have to move back to naming the entry point's file like the name of chunk with `[name].js`. Furthermore you'd have to change the templates where `window.adhocracy4.renderWhatever` is called back to `window.opin.renderWhatever` depending on whether the particular component comes from opin or adhocracy4.

slomo:
replaced by #693

PR #695, opened 2017-02-22
Implement receiver spliting for emails

slomo:
- send each receiver thier own copy of the email


slomo:
Remark: This should be ported to adhcoracy4 at some point and we probably want to send those messages async.

Issue #675, opened 2017-02-16
Map shows multiple (alternative?) worlds


rittermo:
When you zoom the OSM maps out, you can see more than one world map. I don’t dare to set markers on one of the alternative worlds for fear of a black hole or a gate to an alternative world. It happens on all instances of the map widget. Please fix!
![bildschirmfoto 2017-02-16 um 16 31 37](https://cloud.githubusercontent.com/assets/15341015/23027777/ae50292c-f465-11e6-9e14-ddecc5a4d0ab.png)

slomo:
closed with #696

PR #696, opened 2017-02-22
Mn 2017 02 fix map issues

### LINKED ISSUE #620: Maps: Pop up should be placed right onto its pin
When I click on a map pin, the pop up appears slighty above. 
I think it would look and feel much nicer if the pop up overlapped the pin. 

Here's a comparison: 

bildschirmfoto 2017-01-23 um 16 43 04

MagdaN:
fixes #620 #675


slomo:
There were some issues in the project create view (related to the map being in a tab). I fixed them by hard coding zoom level 2 as minimum.

PR #698, opened 2017-02-22
fix the error message problem for maps

MagdaN:


slomo:
I did some small changes. Will still merge myself here.

PR #700, opened 2017-02-23
Mn 2017 02 refactor ewelles changes

MagdaN:
- [ ] login data to flashpoll server needs to be in local setting
- [x] get google maps API Key from local settings
- [x] load highcharts from npm
- [x] move flashpoll-templates to flashpoll app
- [x] move flashpoll template tags to flashpoll app
- [ ] clean up history

google maps does not need to be added via npm, as tiles and js code are downloaded in js code with the api key included (see documentation [here](https://developers.google.com/maps/documentation/javascript/))


MagdaN:
@slomo can you please test and clean up the history?

PR #714, opened 2017-03-03
fix user test

### LINKED ISSUE #709: fix randomly failing user api test

MagdaN:
fixes #709


slomo:
Do you know what the issue was and why this does fix the issue?

MagdaN:
I could not really test this, as the error occurs randomly, but I guess because 'user' was added to the function an additional user was created and this users username sometimes started with 'a'.

slomo:
I see. Then lets roll this out.

PR #715, opened 2017-03-03
add file upload widget

MagdaN:
this is pretty similar to the image upload widget in core, we should check, if this is needed in core also.


slomo:
This behaves a bit differently than the image upload in regards to deleting files.

slomo:
I just saw you discussion with @damusp in zepplin. So I guess we'll merge this first? Will there be an additional story to fix the remaining issues in the upload form?

MagdaN:
What remaining issues do you mean?

slomo:
As far as I understand there should be a delete button similar to the one for the images.

PR #739, opened 2017-03-08
Mn 2017 03 fix button text size

### LINKED ISSUE #678: Title Case in buttons on opin-dev
I see that on opin-dev, we now use title case for buttons. I think this is not ideal, since it is only common for English buttons. In German it creates a lot of problems, for example, because verbs can get mistaken as nouns and so on. Even in English, some words like “and” or “or” are not normally spelled with a capital letter in titles. I think the upper-case spelling was more reasonable for all languages.

bildschirmfoto 2017-02-16 um 17 08 56

MagdaN:
fixes #678


slomo:
I find those commit messages a bit short. Could you explain a bit more what is going on in future commits (e.g. "remove text-transform from button classes" and "change case of button translations string")?

PR #742, opened 2017-03-09
add a detail view for invites

slomo:
- rename old view to update and still require login
 - new view doesn't require login and explains the invite first

This is related to some email issue, but also intended to be an overall improvment for the invite flow.

### DIFF
--- a/euth/memberships/models.py
+++ b/euth/memberships/models.py
@@ -42,7 +42,7 @@ class Invite(base.TimeStampedModel):
 
     def get_absolute_url(self):
         url_kwargs = {'invite_token': self.token}
-        return reverse('membership-invite-accept', kwargs=url_kwargs)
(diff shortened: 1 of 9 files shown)


MagdaN:
It works nicely, but can you explain to me again why this is needed

slomo:
For one it is an improvement to the invite flow to be more accessible for users without accounts. This is because doesn't require you to login/register anymore without knowing why you should login/register. On the other hand, there was some penalty in the SPAM filters for including a link to a page that immediately redirects. Is suspect that is, because it is intransparent to send the user a link, which the user might have "eyeballed" and than redirecting away from it (but I am not sure here, why the SPAM filter didn't like it).

The SPAM penalty is quire minor for this. But I think it also makes now more sense than before from a usability point of view.

PR #748, opened 2017-03-13
Change email wording and rendering

slomo:
-  adopt more structured email format
    - five parts: subject, headline, content, cta (url and label), reason + (fixed footer)
    - reason part explains why the recipient gets this email and how to disable it
    - html emails consist of headline, content, cta (as button) and reason
    - txt mails consist of headline, content, cta (as link) and reason
- can all be written to the same template
- open issues: How to deal with italic and inline links in txt emails (probably we should just filter all tags)


euth_wagtail/templates/email_base.html
```diff
@@ -39,14 +45,43 @@
...
+              <hr />
+              {% trans 'OPIN.me is an online platform for youth participation operated by' %}<br />
+              nexus GmbH, Otto-Suhr-Allee 59, D-10585 Berlin<br />
+              Liquid Democracy e.V., Am Sudhaus 2, D-12053 Berlin<br />
+          </td>
+      </tr>
+      <tr>
+          <td class="footer" style="padding-top: 10px; padding-bottom: 10px;">
+              <a href="{{ email.get_host }}/contact">Contact</a><br />
+              <a href="{{ email.get_host }}/terms-of-use">Terms of use</a>
```
MagdaN: should we add the links to these pages to SETTINGS and only add them if they are in the SETTINGS

MagdaN:
Looks nice, but I am wondering if we should get the contact and term of use url to the settings

PR #762, opened 2017-03-23
Move email sending to a4

slomo:


slomo:
We might consider waiting for https://github.com/liqd/adhocracy4/pull/48 to close.

PR #775, opened 2017-04-10
Removing possibilities for script injection

slomo:
- fixes script injection in documents
  - by passing in data attribute instead of safe marked json
- adapt parameter passing for all other react widgets
- eliminate most embedded scripts
  - left:
    -  flashpoll dashboard result tab (can probably be moved)
    - email management view (should probably be removed anyway)
    - adhocracy4 components (could be changed without much hassle)
  - allows us to change [content security policy](https://en.wikipedia.org/wiki/Content_Security_Policy) as suggest by @mkind 
  - leads to js code being evaluated after DOM parsing, which results in js rendered compontents flickering upon load


euth/contrib/static/js/app.js
```diff
@@ -15,3 +15,23 @@ module.exports = {
   'renderFollow': ReactFollow.renderFollow,
   'renderUserList': ReactUserList.renderUserList
 }
+
+var initilizeWidget = function (project, name, initializer) {
+  if (!initializer) {
+    initializer = name
+    name = project
+    project = 'euth'
+  }
```
xi: Why not have `project` as 3rd parameter?

euth/contrib/static/js/app.js
```diff
@@ -15,3 +15,23 @@ module.exports = {
...
+
+  document.addEventListener('DOMContentLoaded', function () {
```
xi: I am not entirely sure of the differences, but `$(document).ready()` could be more robust. A common case is that it will still execute if the content is already loaded.

euth/contrib/static/js/app.js
```diff
@@ -15,3 +15,23 @@ module.exports = {
   'renderFollow': ReactFollow.renderFollow,
   'renderUserList': ReactUserList.renderUserList
 }
+
+var initilizeWidget = function (project, name, initializer) {
```
xi: typo: initialize (a)

euth/contrib/static/js/app.js
slomo: Because the api is more logical that way. From the context to the widget to the parameters.

euth/contrib/static/js/app.js
slomo: Fixed in a later commit to `initialise`.

euth/contrib/static/js/app.js
slomo: I wanted to avoid a jQuery dependency here. At a quick internet search yielded this as a equivalent vanilla java script option.

euth/flashpoll/templates/euth_flashpoll/includes/project_form_result_flashpoll_tab.html
```diff
@@ -5,8 +5,8 @@
 <script type="text/javascript">
 
 var apiKey = '{% api_key %}';
-var pollresult = {{pollresult|get_simplejson|safe}};
-var poll = {{poll|get_simplejson|safe}};
+var pollresult = JSON.parse('{{ pollresult|get_simplejson|escapejs }}');
+var poll = JSON.parse('{{ poll|get_simplejson|escapejs }}');
```
xi: I am not entirely sure that this is secure. The documentation is a bit unclear. https://stackoverflow.com/questions/23331988/#answer-23341128 suggests that it is, but data attributes are still preferable.

slomo:
@MagdaN I also rewrote the flashpoll stuff. Please rereview and merge if possible.

PR #781, opened 2017-04-12
Fix dashboard permissions and change urls

slomo:


euth/dashboard/views.py
```diff
@@ -38,6 +38,10 @@ def organisation(self):
         if 'organisation_slug' in self.kwargs:
             slug = self.kwargs['organisation_slug']
             return get_object_or_404(org_models.Organisation, slug=slug)
+        if 'slug' in self.kwargs:
+            slug = self.kwargs['project_slug']
```
2e2a: Should be slug (Just saw it doing the same for meinberlin)

slomo: I renamed the slug parameter to `project_slug` so it should be `project_slug`.

PR #784, opened 2017-04-12
Archiving projects

mkind:
* sending `archive` in POST edit request will set a project's `is_archive` flag
* the flag requires that no phase is running anymore
* the project dashboard only shows the `archive` button if project is archivable (not yet archived and no phase left running)
* organisation view does not show archived projects anymore
* archived projects have an `archived` label in project list


slomo:
:exclamation: Rebased onto to current master :exclamation:

PR #787, opened 2017-04-13
Make organisations link configruable

slomo:


slomo:
I had to include the migration for the configurable header to avoid conflicts when merging release2.3 with master.

PR #788, opened 2017-04-19
Validate document on rest api

slomo:


euth/documents/validators.py
```diff
@@ -0,0 +1,16 @@
+from django.core.exceptions import ValidationError
+
+from django.utils.translation import ugettext as _
+
+
+def single_document_per_module(module, pk=None):
+    from .models import Document
```
MagdaN: can this go to the top of the file?

slomo: I tried and it results in an import loop. Do you have any Idea how I could prevent that?

MagdaN: Urgh, well that we have to stick to it (maybe add a comment?)

PR #791, opened 2017-04-20
Fix sending of logo with email

slomo:


MagdaN:
Somehow there is no logo in my email

slomo:
Fixed now, will probably still not work in production due to deployment bug.

PR #792, opened 2017-04-20
Improve momentjs integration

### LINKED ISSUE #673: Minimize moment js weight
Only include locales needed in moment.js (see https://github.com/liqd/a4-meinberlin/blob/master/webpack.config.js#L76)

slomo:
Fixes #673

### DIFF
--- a/euth/contrib/static/js/app.js
+++ b/euth/contrib/static/js/app.js
@@ -8,12 +8,7 @@ var ReactUserList = require('../../../dashboard/static/user_list/react_user_list
 
 require('../../../../euth_wagtail/assets/js/euth_wagtail')
 
-var configureLocale = function (locale) {
-  moment.locale(locale)
-}
-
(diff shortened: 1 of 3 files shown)


MagdaN:
time difference is not formatted the expected:
- it is not translated
- should it not say 'a couple of minutes ago' instead of 'Latest edit on 20. Apr. 2017 13:38'
bildschirmfoto 2017-04-20 um 13 25 46


BUT: I figured out it looks the same on staging (so maybe this is a bug in core?)

slomo:
I think it works. The date in your screen shoot is in the German date format, which proves that moment works. In addition it is [configured](https://github.com/liqd/adhocracy4/blob/master/adhocracy4/comments/static/comments/Comment.jsx#L107) to always output in long format, that is why the relative format is not used. I think that might have been even an requirement.

PR #808, opened 2017-04-27
Archiving projects via projectlist

mkind:
* Adding un-/archiving buttons in dashboard project list instead of project update form
* Archived projects are not editable anymore

### DIFF
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -294,6 +294,7 @@ class ProjectUpdateForm(multiform.MultiModelForm):
                 objects['module_settings'].save()
 
     def clean(self):
+        super().clean()
         objects = super().save(commit=False)
         project = objects['project']
         if project.is_archived:
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -149,6 +149,13 @@ class ProjectForm(forms.ModelForm):
         return formsections
 
 
+class ProjectArchiveForm(forms.ModelForm):
+
+    class Meta:
+        model = project_models.Project
+        fields = ['is_archived']
+
+
 class PhaseForm(forms.ModelForm):
     delete = forms.IntegerField(widget=forms.HiddenInput(), initial=0)
 
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -270,7 +270,7 @@ class ProjectUpdateForm(multiform.MultiModelForm):
                 Offlinephase.objects.get_or_create(phase=new_phase)
 
     def save(self, commit=True):
-
+        self.clean()
         objects = super().save(commit=False)
         project = objects['project']
         module = project.module_set.first()
@@ -292,6 +292,13 @@ class ProjectUpdateForm(multiform.MultiModelForm):
             project.save()
             if 'module_settings' in objects:
                 objects['module_settings'].save()
+
+    def clean(self):
+        objects = super().save(commit=False)
+        project = objects['project']
+        if project.is_archived:
+            raise ValidationError(_('Archived projects are read-only.'))
+        return self.cleaned_data
 
 
 class ProjectCreateForm(multiform.MultiModelForm):
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -297,7 +297,9 @@ class ProjectUpdateForm(multiform.MultiModelForm):
         objects = super().save(commit=False)
         project = objects['project']
         if project.is_archived:
-            raise ValidationError(_('Archived projects are read-only.'))
+            raise ValidationError(
+                    _('Archived projects are read-only.'),
+                    code='read-only')
         return self.cleaned_data
 
 
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -236,15 +236,9 @@ class ProjectUpdateForm(multiform.MultiModelForm):
                 get_module_settings_form(module.settings_instance),
             ))
 
-        no_phase_left = True
-        for phase in qs:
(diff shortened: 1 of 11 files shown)


euth/dashboard/forms.py
```diff
@@ -263,7 +270,7 @@ def _create_phase(self, phase, commit, module):
                 Offlinephase.objects.get_or_create(phase=new_phase)
 
     def save(self, commit=True):
-
+        self.clean()
```
slomo: Why is it necessary to call clean here? Should this be done by the view before calling save()?

euth/dashboard/forms.py
```diff
@@ -282,18 +288,20 @@ def save(self, commit=True):
...
-                no_phase_left = False
-
-        project.is_archived = project.is_archived and no_phase_left
-
         if commit:
             project.save()
             if 'module_settings' in objects:
                 objects['module_settings'].save()
 
+    def clean(self):
```
slomo: you probably want to also to call `super().clean()` just to be

euth/dashboard/views.py
```diff
@@ -120,10 +120,16 @@ class DashboardProjectListView(DashboardBaseMixin,
     filter_set = filters.ArchivedFilter
 
     def get_queryset(self):
-        return super().get_queryset().filter(
+        qs = super().get_queryset().filter(
             organisation=self.organisation
         )
 
+        for project in qs:
```
slomo: The querysets are lazy, by iterating through them here you are forcing the orm to load all project from the data base. Even if they would be filtered at a later stage (eg, by pagination or filters).

There are two other ways to archive the same. Either define is_archivable as a getter (property) on the project. Or use the [annotate](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#annotate) feature of the queryset.

euth/dashboard/templates/euth_dashboard/project_form.html
```diff
@@ -10,6 +11,10 @@ <h1 class="black-divider dashboard-title dashboard-title-no-margin">{{ heading }
         <div class="errorbar">
             {% trans 'Unable to submit form. Please fix the errors.' %}
         </div>
+        {% elif project and project.is_archived %}
```
slomo: You don't need to do such checks in django templates. The template language will them do itself.

euth/dashboard/views.py
```diff
@@ -234,6 +240,34 @@ def get_form_kwargs(self):
...
+
+    def get_success_url(self):
+        return reverse('dashboard-project-list',
+                       kwargs={
+                           'organisation_slug': self.organisation.slug,
+                       })
+
+    def form_valid(self, form):
+        project = form.instance
+        project.is_archived = self.archiving and project.has_finished
```
slomo: I am wondering whether it wouldn't be better to validate in `form.clean` that the project has indeed finished. Also this does save the model twice (once here and once in `UpdateView.form_valid()`.

euth/dashboard/templates/euth_dashboard/project_list.html
```diff
@@ -63,7 +63,22 @@ <h1 class="black-divider dashboard-title">{% trans 'Projects' %}</h1>
...
+                              {% csrf_token %}
+                              <button type="submit" class="btn btn-sm btn-hover-primary" title="{% trans 'archive project' %}">
+                                <i class="fa fa-archive"></i>
+                              </button>
+                            </form>
+                            {% elif project.is_archived %}
+                            <form action="{% url 'dashboard-project-unarchive' project.slug %}" method="post">
+                              {% csrf_token %}
+                              <button type="submit" class="btn btn-sm btn-hover-primary" title="{% trans 'unarchive project' %}">
+                                <i class="fa fa-archive"></i>
```
slomo: You nee to hide this (and the button above) from screen readers (by setting aria-hidden='true'). The scree reader will than use the title attribute instead.

euth/dashboard/views.py
```diff
@@ -234,6 +240,34 @@ def get_form_kwargs(self):
...
+class DashboardProjectArchiveView(DashboardBaseMixin,
+                                  rules_views.PermissionRequiredMixin,
+                                  generic.UpdateView):
+    model = project_models.Project
+    form_class = forms.ProjectArchiveForm
+    success_message = _('Project has been archived.')
+    permission_required = 'euth_organisations.initiate_project'
+    slug_url_kwarg = 'project_slug'
+    menu_item = 'project'
+    archiving = False
```
slomo: I had to look twice to figure out how this works :smile:

euth/dashboard/views.py
mkind: > I am wondering whether it wouldn't be better to validate in form.clean that the project has indeed finished

I like the idea. Since the `form.clean` requires raising an exception and the current templates do not take exceptions into account, I would prefer `project.is_archived = self.archiving and project.has_finished`.

euth/dashboard/forms.py
mkind: Done in https://github.com/liqd/a4-opin/commit/a44b1c2704557bc776aff76081e8c4cca0e2d904.

euth/dashboard/templates/euth_dashboard/project_list.html
mkind: Fixed in https://github.com/liqd/a4-opin/commit/7778bb1e0bb48602e83b859375f9d64c3fc89484.

euth/dashboard/views.py
mkind: Very good point. I fixed it by using `@property`. See the [corresponding a4 pull request](https://github.com/liqd/adhocracy4/pull/93) and https://github.com/liqd/a4-opin/commit/92f3401ec1e2e03db37cc299c735fc2bf5db50b8.

PR #811, opened 2017-05-08
added georgian language

fuzzylogic2000:


slomo:
@fuzzylogic2000 Now you also have to add gregorian to transifex. I can help you with that.

PR #820, opened 2017-05-16
Ka 2017 05 user timezone

fuzzylogic2000:
- added timezone to model
- added field to chose to user dashboard
- implemented middleware
- show used timezone in project view


euth/contrib/middleware.py
```diff
@@ -0,0 +1,12 @@
+import pytz
+
+from django.utils import timezone
+
+
+class TimezoneMiddleware(object):
```
slomo: Extending a class from `object` is the Python2 way to tell the interpreter, that you want new style classes. While it doesn't harm in Python3 it is definitely not needed anymore (see [stackoverflow](http://stackoverflow.com/questions/15374857/should-all-python-classes-extend-object))

euth/documents/validators.py
```diff
@@ -1,5 +1,4 @@
 from django.core.exceptions import ValidationError
-
```
slomo: You commit a lot of those changes in commits that try to implement a specific feature. I guess because your run isort. While this also does no harm it is cleaner to put this in an extra commit, or even extra branch. Because right now your changes to the user profile touched files in all components.

euth/users/models.py
```diff
@@ -108,7 +109,7 @@ class User(auth_models.AbstractBaseUser, auth_models.PermissionsMixin):
     gender = models.CharField(
         blank=True,
         verbose_name=_('Gender'),
-        max_length=1,
+        max_length=2,
```
slomo: Nice :+1: good catch.

Issue #644, opened 2017-01-30
refactor offline phase dashboard js


MagdaN:
1) ~~check if transition between local time in FE and UTC in BE is working~~ (moved to #827)
2) use empty_form https://docs.djangoproject.com/en/1.10/topics/forms/formsets/#empty-form

fuzzylogic2000:
In ka-2017-05-user-timezone-datepicker, there are no fields to insert a date when I create the offline phase. but only when I edit it.

hello-kathi:
@phillimorland Don't know what that is :) Please lable with reminder if it is important or close.

phillimorland:
seems to work now

PR #830, opened 2017-05-19
Use empty_form for adding offline phase

slomo:
solves #644

- [x] use formset.emtpy_form to render html for offline form
- [ ] ensure that the form template renders as complete offline form
     - remove post parse modifications from `updateNewElement`
- [ ] use can_order feature of formset to specify to order in the formset
    - remove replace `setPhasesId` with `updateOrder`
- [ ] remove hidden type field since it basically allows the user to modify the project to any sequence of phases that they want
- [ ] NTH: Factor this out into reusable component for each formset

### DIFF
--- a/euth/dashboard/static/add_offline_phase/add_offline_phase.js
+++ b/euth/dashboard/static/add_offline_phase/add_offline_phase.js
@@ -2,8 +2,8 @@
 window.jQuery(document).ready(function () {
   function updateNewElement (element) {
     element.addClass('phaseform-offline-phase')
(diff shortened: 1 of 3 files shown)


slomo:
@MagdaN What do you think about the changes and the step left that are outlined above?

slomo:
@MagdaN Could you please comment here, if it is worth continuing.

MagdaN:
Maybe we should wait and see if we really can reuse this.

slomo:
Closing for now, because it is probably deprecated by the new dashboard?

PR #837, opened 2017-05-24
Mkind 2017 05 dst frontend

mkind:
* cleans up the DST templating by moving different things in different files
* introduces second fine tuning step in DST
* introduces time needed in hours per week in DST result
* fixes typo in complexity calculation


slomo:
I added a single commit renaming `finetuning 2` into `feasibility`.

slomo:
LGTM

PR #838, opened 2017-05-24
browser timezone is added to user profile at signup

fuzzylogic2000:


euth/users/static/users/js/user_timezone.js
```diff
@@ -0,0 +1,6 @@
+var $ = window.jQuery
```
slomo: You probably want to handle this type of renaming by using an [immediately invoked function expression](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression) (iife). This is definitely not a server case, but what you are doing here is creating a global variable in the global namespace (called `$`). With an iife it would look like:
```
(function ($) {
   // your code here
})(window.jQuery)
```

home/models.py
```diff
@@ -59,6 +58,7 @@ class RSSImport(models.Model):
     def __str__(self):
         return self.rss_title_en
 
+
```
slomo: Probably would be better off in a Gardening commit :smile: (but nevermind).

euth/users/models.py
```diff
@@ -197,6 +197,7 @@ def signup(self, username, email, timezone, commit=True):
         """Update the fields required for sign-up."""
         self.username = username
         self.email = email
-        self.timezone = timezone
+        if timezone in common_timezones:
```
mkind: To me this commit seems superfluous. AFAIK django validates the form input against the model. @slomo do you have an opion about this?

slomo: I would agree. Could this be a check against `None`?

home/models.py
fuzzylogic2000: It keeps happening to me in Opin, that the when I try to commit, the linter complains about files, I never touched. So I just fix them. Should I check for that before I start working?

euth/users/models.py
mkind: The worst case assignment seems to be an empty string. If this is not desired one should not set `blank=True`.

home/models.py
slomo: Best practice is to commit them in an extra commit with message `Gardening` or similar. You can do it before or after your changes.

We could enable linting on travis to catch this early on.

PR #842, opened 2017-05-29
Mkind 2017 05 fixing dst no hour bug

mkind:
* Fixes a bug found by @SabineJulia that was trigged by using the fallback DST result. For the fallback there was no `timeneeded` calculation.  
* adding regression test


slomo:
Good catch @SabineJulia

PR #870, opened 2017-06-21
changed the color of the rating bottons 

ahmedjenbaz:
changed the color of the rating bottons to gray when the rating is not active in the current phase

### DIFF
--- a/euth_wagtail/assets/scss/components/_rating.scss
+++ b/euth_wagtail/assets/scss/components/_rating.scss
@@ -37,11 +37,11 @@
 
     &:disabled {
         background: transparent;
-        border: none;
-        color: $brand-success;
+        border-color: $gray;
+        color: $gray;
 
         i {
-            color: $brand-success;
+            color: $gray;
         }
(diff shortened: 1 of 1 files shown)


euth/ideas/templates/euth_ideas/includes/idea_list_tile.html
```diff
@@ -11,7 +11,7 @@
         <div class="row">
             <div class="col-xs-12 col-sm-3 col-sm-push-9 idea-meta">
                 <span class="idea-upvotes idea-meta-item">{{ idea.positive_rating_count }} <i class="fa fa-chevron-up" aria-hidden="true"></i></span>
-                <span class="idea-downvotes idea-meta-item">{{ idea.negative_rating_count }} <i class="fa fa-chevron-down" aria-hidden="true"></i></span>
```
slomo: This looks pretty much like an accidental delete.

slomo:
I fixed the error myself, because it was rather complicated.

Issue #431, opened 2016-11-01
Text Proposal for "Accessible for all registered users on OPIN"


NadineKarbach:
Hey a-team, i've been wondering whether the explanation text is easy to understand for our users and initiators. Here, i may propose an updated wording, ready for discussion:
"All projects are open to public: All registered users and everyone else can see and participate in your project. You can chose to set it "private" by unchecking the box below. In private projects you can invite participants yourself and participants can ask for access. Teasers for your project including title and short description will always be visible to everyone though."
What do you think @slomo @MagdaN @rittermo @vellip @damusp @IsabelSensan ?

![bildschirmfoto 2016-11-01 um 18 08 54](https://cloud.githubusercontent.com/assets/13625047/19899496/abc5779c-a060-11e6-9640-75e571956ab5.png)

damusp:
Hey Nadine, 
I think your wording update is helpful, but what actually bothers me here is the checkbox. 
What I don’t like is that the user has to uncheck the box if he wants to set the project to private. Doesn't quite feel right.

I believe that, from a user perspective, radio buttons would be way easier to understand in this case.

vellip:
We discussed this earlier today, and I actually prefer the checkbox, but I
would make the label (which comes after the checkbox) clearer. The checkbox
seems just simpler, and therefore easier to grasp, because it's just one
option instead of two. With a clear label (see below) it would also be
obvious what the other possibility is.

I'd prefer a label like "This project is private, participants have to
request access or be invited." And by default, the checkbox is *not*
checked.

https://uxplanet.org/radio-buttons-ux-design-588e5c0a50dc#.j3xjm6ey9

slomo:
I am with @damusp and would favour a radio button. Since than both cases can be made clear to the user.

   - [ ] all registered opin users
   - [ ] only users invited by you

The text above should also state, that this is not only about participating but also about seeing the projects content.

slomo:
@vellip I also thanks for the link, I think it makes a strong case for radio buttons :smile:

NadineKarbach:
@slomo @vellip @damusp Thanks for the link! Great help in understanding the radio button and the UX philosophy behind it! :-)

MagdaN:
@slomo @damusp what is the solution here?

damusp:
I'm still for radio buttons. 
Also, the labels @slomo suggested earlier fit really well imo.

mkind:
[...]

PR #874, opened 2017-06-27
Ah 2017 6 privacy botton

ahmedjenbaz:
fixed isse #431

### DIFF
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -127,14 +127,6 @@ class ProjectForm(forms.ModelForm):
         self.instance.is_draft = 'save_draft' in self.data
         return super().save(commit)
 
-    def get_checkbox_label(self, name):
-        checkbox_labels = {
-            'is_public': _('Accessible to all registered users of OPIN.me')
-        }
-        if name in checkbox_labels:
-            return checkbox_labels[name]
-        else:
-            return ''
 
     @property
     def formsections(self):
--- a/euth/dashboard/forms.py
+++ b/euth/dashboard/forms.py
@@ -115,9 +115,14 @@ class ProjectForm(forms.ModelForm):
         model = project_models.Project
         fields = ['name', 'description', 'image', 'information', 'is_public',
                   'result']
-        widgets = {'is_public': RadioSelect(choices=[
-                (True, _('Public')),
-                (False, _('Private'))])}
+        widgets = {
+            'is_public': RadioSelect(
+                choices=[
+                    (True, _('All users can participate (public).')),
+                    (False, _('Only invited users can participate (private).'))
+                ]
+            )
+        }
 
     def save(self, commit=True):
         # calling flashpoll service
@@ -126,7 +131,6 @@ class ProjectForm(forms.ModelForm):
 
         self.instance.is_draft = 'save_draft' in self.data
         return super().save(commit)
-
 
     @property
     def formsections(self):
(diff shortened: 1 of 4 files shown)


euth/dashboard/forms.py
```diff
@@ -114,7 +116,11 @@ class ProjectForm(forms.ModelForm):
     class Meta:
         model = project_models.Project
         fields = ['name', 'description', 'image', 'information', 'is_public',
-                  'result']
+        'result']
+        widgets = {'is_public': RadioSelect(choices=[
+            (True, 'Private'),
```
mkind: Good work @ahmedjenbaz :smile: The translatable version ` _('Private')` might be better here.

euth/dashboard/forms.py
```diff
@@ -115,6 +115,9 @@ class Meta:
         model = project_models.Project
         fields = ['name', 'description', 'image', 'information', 'is_public',
                   'result']
+        widgets = {'is_public': RadioSelect(choices=[
+                (True, _('Private')),
+                (False, _('Not Private'))])}
```
fuzzylogic2000: 'Not Private' should be 'Public', and the booleans should be the other way round. (As the model field is called is_public, the field should be set true, when the project is public and false, when it is private.)

euth/dashboard/templates/euth_dashboard/includes/project_form_information_tab.html
```diff
@@ -20,11 +20,9 @@ <h3 class="form-section-heading">{{ name }}</h3>
                 {{ field.label }} {% if field.field.required %}*{% endif %}
             </label>
             <p class='form-help-text'>{{ field.help_text }}</p>
-            {% if field|widget_type == 'checkboxinput' %}
+            {% if field|widget_type == 'radioselect' %}
             <label>
-                {% get_checkbox_label form.project field.name as checkboxlabel %}
```
fuzzylogic2000: we still need to check if the tamplate tag is needed elsewhere.

euth/dashboard/templates/euth_dashboard/includes/project_form_information_tab.html
```diff
@@ -20,11 +20,9 @@ <h3 class="form-section-heading">{{ name }}</h3>
                 {{ field.label }} {% if field.field.required %}*{% endif %}
             </label>
             <p class='form-help-text'>{{ field.help_text }}</p>
-            {% if field|widget_type == 'checkboxinput' %}
+            {% if field|widget_type == 'radioselect' %}
```
fuzzylogic2000: do we still need this if for the styling? and the <label>?

euth/dashboard/templates/euth_dashboard/includes/project_form_information_tab.html
```diff
@@ -20,14 +20,10 @@ <h3 class="form-section-heading">{{ name }}</h3>
                 {{ field.label }} {% if field.field.required %}*{% endif %}
             </label>
             <p class='form-help-text'>{{ field.help_text }}</p>
-            {% if field|widget_type == 'checkboxinput' %}
-            <label>
-                {% get_checkbox_label form.project field.name as checkboxlabel %}
```
fuzzylogic2000: The assignment-tag get_checkbox_label and the function it uses can and should be deleted, as these are not used in any other template.

slomo: I think this if in the tender logic (that checks the widget_type) could also be removed then.

PR #881, opened 2017-07-19
Mkind 2017 07 introducing categories

mkind:
* adding categories to dashboard
* adding categories to proposal
* adding tests


MagdaN:
LGTM just two small things

-  maybe add a sentence like 'No proposals found that match this category' to proposal list

- when adding a new category in the dashboard the page jumps up, due to some or missing anchor link (I guess), maybe that could be fixed?

PR #884, opened 2017-07-20
Removed whitespaces and added result tab for end users

ewelle:
In this pull request:
    - we add result tab for end users
    - we clean the code from whitespaces


ewelle:
@MagdaN ,
Yes, this is an issue we need to investigate. And It will be resolved with the full implementation of private processes. 
In fact on the embed flashpoll webapp (opin.flashpoll.eu), we use sessionId to identify users, So that, if you have already answered to a poll, you can't answer it anymore. You can just see the results. That's why you are automatically redirected to the results tab.
So I think in your case, you used the same session twice to answer to the poll. This will be corrected with the full implementation of private processes. But in the mean time, when switching users, you need to delete the cookies on the opin.flashpoll.eu as well.

slomo:
After the merge with master there is no more code in this PR, therefore closing.

PR #885, opened 2017-07-20
Added is private field to poll

ewelle:
In this pull request:
    - we add private processes management for the flashpoll service


ewelle:
- For the first point concerning date format, I will correct this very soon.

- For the second, the implementation of private processes is not done yet. This is just a first step toward that. Taging polls as private before pushing them to flashpoll service. At the current stage the private polls are just hidden from the poll list in the flashpoll.eu, but of course if you have the poll url, you can directly access to from any browser.  This is one of the points I wanted to talk to you about. "How to handel the private processes".
.
-  The download button is only available for admins on the result tab of the poll editor. End users can't download poll results.

euth/flashpoll/services.py
```diff
@@ -26,25 +25,32 @@ def send_to_flashpoll(data):
...
+            jsonGenerator['descriptionMediaURLs'] = [""]
             jsonGenerator['descriptionMediaURLs'] = [""]
             jsonGenerator['keywords'] = []
             jsonGenerator['resultVisibility'] = 0
             jsonGenerator['startTime'] = startTime
             jsonGenerator['endTime'] = endTime
             jsonGenerator['preview'] = 'save_draft' not in data
+            # isPrivate
+            if 'project-is_public' in data:
+                if data['project-is_public'] == "on":
```
slomo: This can be queried from `objects['project'].is_private`.

slomo:
There are no more change in this branch therefor close.

PR #889, opened 2017-07-25
Mn 2017 07 export ideas

MagdaN:
First version for exporting ideas as xls, should be easy adaptable for documents. However if we want to make this really generic, we would probably need a custom url scheme for exports, where you pass in the module slug and then the view takes care of getting the right queryset. We can check with the other team - they loop through all phases and if within the phases view an explicit export view is set they use that one (I think)

Also, I am not sure about the permissions, who is allowed to pull that data?


MagdaN:
I am using the code from the other team now - if we decide to move this to A4 we could also add the tests that they have written

euth/contrib/exports.py
```diff
@@ -0,0 +1,247 @@
...
+        for rownum, row in enumerate(self.export_rows(), start=1):
+            for col, field in enumerate(row):
+                worksheet.write(rownum, col, field)
+
+        workbook.close()
+
+        return response
+
+
+class VirtualFieldMixin:
```
MagdaN: Yes, I copied it from the other team, I guess its about FK-Fields

euth/contrib/exports.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def get_queryset(self):
+        return super().get_queryset().filter(module=self.module)
+
+    def get_header(self):
+        return self._header
+
+    def export_rows(self):
+        for item in self.get_queryset().all():
+            yield [self.get_field_data(item, name) for name in self._names]
```
MagdaN: yes

euth/ideas/templatetags/idea_tags.py
```diff
@@ -11,3 +13,8 @@ def get_range(number, listcount):
         return range(listcount-4, listcount+1)
     else:
         return range(number-2, number+3)
+
+
+@register.assignment_tag
+def is_idea_list(module):
+    return Idea.objects.filter(module=module).count() > 0
```
MagdaN: but I would also only add an download button if there are any ideas to download

slomo:
Okay, so I shall merge this branch now?

MagdaN:
yes if you like

Issue #901, opened 2017-08-03
category filter and sorting on the same level


gereonrahnfeld:
If possible it would be great if we had category filter and sorting on the same level. Otherwise there is a big gap between them. Doesn't look good. 

This is how it looks when there are several proposals with the same category:

![filter and sorting on same level](https://user-images.githubusercontent.com/15954895/28911495-0b8d7c4e-7831-11e7-8781-c89cb9608da7.png)

And this is how it looks when there is only one proposal:

![bildschirmfoto 2017-08-03 um 09 50 07](https://user-images.githubusercontent.com/15954895/28911543-2bddca3a-7831-11e7-9226-4e9d4d4fb12f.png)

PR #905, opened 2017-08-03
fixing issue #901

mkind:
* fixing issue #901


MagdaN:
When only one item is displayed the dropdown for filtering by category sticks to the first list item
bildschirmfoto 2017-08-03 um 16 48 03

PR #916, opened 2017-08-16
Mkind 2017 08 dst infopage

mkind:
* introducing blueprint infopages 
* introducing blueprints.names


mkind:
@slomo, @fuzzylogic2000 or @MagdaN, although the tests stil fail (migrations seem to be the reason for failure, so a fix might be easy.), I would appreciate that you have a look at the current state.

PR #917, opened 2017-08-17
added migrations

fuzzylogic2000:
There were some migrations missing. But this seems strange. Did we alter/delete and such stuff four models?

### DIFF
--- a/euth/maps/migrations/0005_move_model_change_point_delete_areasettings.py
+++ b/euth/maps/migrations/0005_move_model_change_point_delete_areasettings.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import adhocracy4.maps.fields
+
+
+class Migration(migrations.Migration):
+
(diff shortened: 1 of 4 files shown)


euth/maps/migrations/0005_move_model_change_point_delete_areasettings.py
```diff
@@ -0,0 +1,27 @@
...
+import adhocracy4.maps.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('euth_maps', '0004_move_area_settings_to_a4'),
+    ]
+
+    operations = [
```
slomo: @MagdaN Do you know what is going on here? Those migrations look a bit scary.

MagdaN: @slomo Yes, thats what we thought.

slomo: Okay I had an other look and this actually looks okay, because we did migrate the data of areasettings to a4_areasettings earlier.

PR #919, opened 2017-08-21
change text

### LINKED ISSUE #918: DST: change a few bits of content
some text within the dst still have to be changed.

Set the agende of an event, a process, a project etc has to be replaced by:
"Learn about what people like most"

bildschirmfoto 2017-08-21 um 12 32 40

there is also one answer choice missing in the fine tuning. Instead of having four answer choices for the third question there should be five: "Don't know"

bildschirmfoto 2017-08-21 um 12 36 02

same counts for the fourth question in feasibility. there should be a fifth answer choise saying "Don't know" as well.

bildschirmfoto 2017-08-21 um 12 44 39

MagdaN:
fixes #918 partly

@slomo I faced problems adding the 'I don't know' option. In code it is there, e.g. [here](https://github.com/liqd/a4-opin/blob/master/euth/blueprints/blueprints.py#L98) but it is not being displayed in the frontend as option. What is the reason for this? I could not figure it out? Is there a problem with having two options with the same value?

### DIFF
--- a/euth/blueprints/blueprints.py
+++ b/euth/blueprints/blueprints.py
@@ -11,6 +11,22 @@ from euth.maps import phases as map_phases
 from .names import BlueprintNames
 
 
+class BlueprintEnum(Enum):
+    def __new__(cls, value, label):
+        obj = object.__new__(cls)
+        obj._value = value
+        obj.label = label
+        return obj
+
+    @property
+    def value(self):
+        return self._value
+
+    @classmethod
+    def get(cls, value):
+        return next(m for m in cls if m._value == value)
+
+
 @unique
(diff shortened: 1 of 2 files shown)


mkind:
@magda Enums with the same value become an [alias for the first label](https://stackoverflow.com/questions/31537316/python-enums-with-duplicate-values/31537386#31537386). So far, there seems to be no clever workaround.

MagdaN:
@mkind thanks for the explanation!

slomo:
Can be solved by iterating over [`enum.__members__`](https://docs.python.org/3.4/library/enum.html#iteration) instead the `enum` itself.

PR #930, opened 2017-08-24
Make blueprint result tiles the same height

slomo:


MagdaN:
Now there are only two blueprints in one row and the images are not centered:
bildschirmfoto 2017-08-24 um 15 01 24

PR #962, opened 2017-10-16
Ka 2017 10 normal user

fuzzylogic2000:


fuzzylogic2000:
needs https://github.com/liqd/adhocracy4/pull/198

PR #989, opened 2017-11-24
add filters to organisation page

MagdaN:


euth/organisations/filters.py
```diff
@@ -0,0 +1,71 @@
...
+            ('-created', 'newest'),
+            ('translations__title', 'title'),
+        ),
+        choices=ORDERING_CHOICES,
+        empty_label=None,
+        widget=OrderingFilterWidget
+    )
+
+    class Meta:
+        model = Project
```
mkind: Is this correct? Shouldn't the model be `Organisation`?

mkind:
Looks good so far. Just one question.

euth/organisations/filters.py
MagdaN: Yes, it should be Organisation, I am not sure why this still works

PR #995, opened 2017-11-27
Sort countries in country dropdown by the current language's alphabet

MagdaN:
looks hacky to me, is there a better way to achieve this?


MagdaN:
I have problem with installing ICU, @slomo please let me know, when you know, if we can also achieve the sorting with our current setup without ICU.

PR #1008, opened 2017-12-04
Enable cache busting

slomo:
Requires an a4 commit to be merged: https://github.com/liqd/adhocracy4/pull/219


slomo:
Can be merged after #1044 is merged.

mkind:
It might be a good idea to fix issue #1045 before merging.

PR #1029, opened 2017-12-11
Remove moment.js since it no longer required by a4

slomo:
Moment was removed here https://github.com/liqd/adhocracy4/pull/154

before:
```
vendor.js     367 kB       2  [emitted]  [big]  vendor
```

after:
```
vendor.js     300 kB       2  [emitted]  [big]  vendor
```


mkind:
yes man, kudos for saving bytes :+1:

PR #1057, opened 2018-01-07
reading statics from bundles dir

mkind:
django failed to read statics from `static/bundles`. added the corresponding dir to config.


MagdaN:
LGTM

PR #1076, opened 2018-01-29
introducing django csp

mkind:
* [x] introducing django csp
* [ ] removing inline script


mkind:
I am not sure, whether there is still inline js missing. The browser reports blocking inline js, but I fail to identify the actual script so far.

MagdaN:
Lets close this until we are working on this again

PR #1091, opened 2018-02-09
adding security parameter

mkind:


kleingeist:
works fine at https://github.com/liqd/a4-meinberlin/pull/1170 so may be merged here, too

PR #1137, opened 2018-03-02
Mn 2018 add offlineevents

MagdaN:
- migrates offlinephases to offlineevents but keeping the models (needs cleanup later on)
- adds CRUD for new model in dashboar


euth/dashboard/templatetags/form_tags.py
```diff
@@ -0,0 +1,36 @@
+from django import template
+from django.conf import settings
+
+from adhocracy4 import phases
+
+register = template.Library()
+
+
+@register.assignment_tag
```
fuzzylogic2000: Assignment tags are deprecated since 1.9, simple tags can now store stuff in variables. Or do we need backwards compatibility here?

PR #1159, opened 2018-03-20
add missing migration

fuzzylogic2000:
I always got a warning that there was a missing migration. @MagdaN Should we add it? Or wait until the offlinephase migrations are obsolete anyway?


fuzzylogic2000:
Ok! I will close this then!

PR #1167, opened 2018-03-21
Use core offlineevents 

kleingeist:
depends on liqd/adhocracy4#248

Note that the migration script for offlinephases has not been tested yet.


MagdaN:
I close this as we are not using offline events in a4, yet.

PR #1198, opened 2018-04-18
unique factory usernames and emails

ihucos:
had to edit random file for style checks to pass (?)


fuzzylogic2000:
Oh, and for the random file: someone smuggled it by the linter. Happens from time to time. If you find something like that, you can only commit the fixed file with a message like 'gardening', so that we can look at your changes without also seeing the changes, you added for the linter.

fuzzylogic2000:
Well, the smuggling doesn't happen that often, but you could still ad it. Or I will! And also add the test for missing migrations.

PR #1268, opened 2018-06-26
Ka 2018 06 remove flashpoll app

fuzzylogic2000:
And number 2! Only merge (and directly deploy?!) after #1267 was merged and deployed everywhere.


fuzzylogic2000:
Falshpoll is still in here: https://github.com/liqd/a4-opin/blob/6ceaa9eeb02082a230d77ac46832e6b005667c67/home/migrations/0029_add_fields_for_example_projects.py#L38
What do we do there? Do we need to also replace it by the other poll?

PR #1271, opened 2018-06-27
Pm 2018 06 spacing fixes#1135

### LINKED ISSUE #1135: account settings: spacing
too much space on the first page of account settings between the first two categories.

 
screen shot 2018-03-01 at 11 50 17

phillimorland:
fixes #1135 

Clicked through site and the reduced margins look good everywhere I checked. 
When there are several margins collected onto 1 section is there a better way to deal with it?


phillimorland:
First 3 commits were intended for the datepicker branch, have undone them in last commit.

Issue #1190, opened 2018-04-17
Usability: rating result looks clickable


ihucos:
![screenshot from 2018-04-17 17-00-08](https://user-images.githubusercontent.com/2066372/38878860-3d912cd0-4262-11e8-9d11-b56450bc78b7.png)

One possible solution:
Make the whole element clickable and with a different (subtle) background color on mouseover

fuzzylogic2000:
Yes! Also annoys me every time!
Do you think it should be clicakble, even if the description of the idea was not read before rating?

ihucos:
>  Do you think it should be clicakble, even if the description of the idea was not read before rating?

Sorry, just read it now. Not sure. I'd say no.

phillimorland:
How about making it clickable through to the idea? they can then easily leave a rating if that was their intention and they have to at least briefly look at the idea.

fuzzylogic2000:
You mean linking it to the idea? So, if the rating in the list is clicked, the idea detail page is openend. I am not sure about that. 
Maybe it's more a design-thing at should look more it's showing a result and less like a button?
What do you think, @gereonrahnfeld or @SabineJulia ?

phillimorland:
We can also try it like we have in meinberlin, the numbers and arrows being the same colour stops the arrows from being highlighted so I think it doesn't look as inviting to click

![opin2](https://user-images.githubusercontent.com/5871230/42030502-4b30b6ae-7ad3-11e8-8098-0add261cf56a.png)

gereonrahnfeld:
But on meinBerlin it is the same than in OPIN isn't it?:
https://meinberlin-dev.liqd.net/projects/laermorte/?mode=list

![screen shot 2018-06-28 at 16 36 34](https://user-images.githubusercontent.com/15954895/42041123-83f24f7a-7af1-11e8-84ec-13c160e594e1.png)

phillimorland:
No, Opin is currently like the top image where the numbers are black so arrows look highlighted.

gereonrahnfeld:
I agree. But on meinberlin-dev the numbers are highlighted as well. This is why I am asking..

fuzzylogic2000:
@ggr On meinberlin the arrows and numbers have the same colours, while on Opin the numbers are black and the arrows are coloured and by that look somehow actionable. I like @phillimorland idea to make both the same colour! What do you say, @SabineJulia ?

gereonrahnfeld:
[...]

PR #1272, opened 2018-06-27
wrapping rating in link to click through to idea

phillimorland:
attempt to resolve usability issue 1190 https://github.com/liqd/a4-opin/issues/1190


MagdaN:
@fuzzylogic2000 what do you think of this solution? I am not sure ...

fuzzylogic2000:
@MagdaN The discussion is happening here: #1190 But stalled. We should finish that!

fuzzylogic2000:
Ok, and discussion decided on other way to go. I will close this.

PR #1278, opened 2018-06-28
Pm 2018 06 accordion issue1096

phillimorland:
I have styled it to resemble what we have in MeinBerlin, is there anything I should have done differently? 
fixes#1096


MagdaN:
I like it that way

PR #1430, opened 2019-03-05
package.json: Update babel

treba123:
That should allow us to use ES6 classes \o/


fuzzylogic2000:
What state is this in? @MagdaN Did you already update babel? @treba123 Does this include your learnings? Does it even work with bootstrap 3?

treba123:
WIP, have to adopt the changes from mB. Babel is not updated yet

treba123:
@fuzzylogic2000 @MagdaN this is ready now

PR #1502, opened 2019-06-04
Mn 2019 06 update django and wagtail

MagdaN:


MagdaN:
@rmader @fuzzylogic2000 I am having real problems to get the tests running, I fixed some of them by running `py.test --no-migrations` which creates the test database based on the current models with out running through all migrations. Some problems come from django-countries which work ok with django 2.1 but not 2.2.2

MagdaN:
@rmader @fuzzylogic2000 ok, I made most tests work, there still is the need to run the tests with `--no-migrations` (otherwise no tables are added to my testdatabase). The ones that I could not fix I commented out

PR #1516, opened 2019-06-13
top menu dropdown: adding correct filepath for timeline_popover

phillimorland:
fixes problem with dropdown js being blocked


rmader:
Well yeah, this looks like another case of js not going through webpack I guess. Looks good.

PR #1525, opened 2019-07-03
Mn 2019 07 fix add moderators

### LINKED ISSUE #1505: Adding moderators: some user names do not show up
When I want to invite specific users as moderators it doesn't work. This counts for admin accounts (e.g. user name 'ggr'). And it also seems as if it is not the user names that is searched for but the email adress or rather the name of the email address before the '@' - sign.

MagdaN:
fixes #1505


rmader:
Ah that's what typeahead is for :) Looks quite obviously right

PR #1799, opened 2020-04-03
Ka 2020 04 translate org fields

fuzzylogic2000:
There are long explanations in the commits. Please read them separately.

### DIFF
--- a/euth/dashboard/static/language_switch/react_language_switch.jsx
+++ b/euth/dashboard/static/language_switch/react_language_switch.jsx
@@ -1,7 +1,7 @@
 var PropTypes = require('prop-types')
 var React = require('react')
 var ReactDOM = require('react-dom')
-var $ = require('jquery')
+// var $ = require('jquery')
 
 class LanguageSwitch extends React.Component {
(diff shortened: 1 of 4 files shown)


euth/dashboard/static/language_switch/react_language_switch.jsx
```diff
@@ -70,10 +66,29 @@ class LanguageSwitch extends React.Component {
           </button>
           <ul className="dropdown-menu">
             {
-              this.props.languages.map(languageCode => {
+              this.props.languages.map((languageCode, i) => {
+                return (
+                  <span key={languageCode}>
```
fuzzylogic2000, line 71: @phillimorland I am a bit unhappy with returning an empty span for the languages that shouldn't be in the list, but couldn't think of a better way. (Apart from also having a list of the inactive languages in the state, but that would be a bit annoying to always update two lists, then.)

phillimorland: i think this is ok for now to be honest

PR #1803, opened 2020-04-08
update dependencies: add @babel/compat-data to rm proposal-numeric-se…

phillimorland:
…parator error


rmader:
He, was about to comment on this, too :)

Don't quite understand why we need this, but if it fixes the issue, let's do it!

PR #1828, opened 2020-04-22
Pm 2020 04 async comments

phillimorland:
initial adding comments, not well styled but will wait for designs before sinking time into it


phillimorland:
@fuzzylogic2000 loading now fixed, the user image I didn't bother with as its a design thing so figured I would wait for that decision?

Rineee:
@phillimorland The editing of comments doesn't work? But deleting is possible.

fuzzylogic2000:
@Rineee As a user? Or admin?

Rineee:
> @Rineee As a user? Or admin?

As an admin.

Rineee:
> > @Rineee As a user? Or admin?
> 
> As an admin.

No actually both admin and user :)

PR #1849, opened 2020-04-28
a4polls/poll_dashboard: adding id and new a4 hash

phillimorland:
…ger reload - not working

@fuzzylogic2000 @rmader at some point I could do with a hand with this but probably not today

### DIFF
--- a/euth_wagtail/templates/a4dashboard/base_dashboard_project.html
+++ b/euth_wagtail/templates/a4dashboard/base_dashboard_project.html
@@ -4,7 +4,7 @@
 {% block dashboard_content %}
 <div class="dashboard-content">
     <div class="row">
-        <div class="col-md-3 dashboard-left-nav">
+        <div class="col-md-3 dashboard-left-nav js-selector-update-dashboard">
             <h2 class="dashboard-left-nav-header">{% trans 'Project' %}: {{ project.name }}</h2>
             <div class="list-group">
                 {% for item in dashboard_menu.project %}
@@ -38,7 +38,7 @@
             {% block dashboard_project_content %}{% endblock %}
         </div>
 
-        <div class="col-md-3">
+        <div class="col-md-3 js-selector-update-dashboard">
             {% include "a4dashboard/includes/progress.html" with value=project_progress.valid max=project_progress.required project=project %}
             {% include "a4dashboard/includes/preview.html" with project=project %}
         </div>
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@babel/preset-react": "7.9.4",
     "@babel/runtime": "7.9.2",
     "@fortawesome/fontawesome-free": "5.13.0",
-    "adhocracy4": "github:liqd/adhocracy4#07a2f63e2fdb21960b8451c074f1e8bd1bef814d",
+    "adhocracy4": "github:liqd/adhocracy4#698cdeba66e10ba204866c93f31c115c50a2790e",
     "autoprefixer": "9.7.6",
     "axios": "0.19.2",
     "babel-loader": "8.1.0",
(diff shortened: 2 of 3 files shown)


euth_wagtail/templates/a4polls/poll_dashboard.html
```diff
@@ -0,0 +1,9 @@
+{% extends "a4dashboard/base_dashboard_project.html" %}
+{% load i18n react_polls %}
+
+{% block dashboard_project_content %}
+    <h1 class="u-first-heading">{% trans 'Edit poll' %}</h1>
+
+    {% react_poll_form poll True %}
```
fuzzylogic2000, line 7: I actually have no idea. The True should reload it, and it's also included like that in a4, soooo, hmmm.

phillimorland: well good to know that's what you meant! :) it's included twice in mB, but that is a different structure, i feel like this is a morning brain thing so will do bootstrap issues for rest of day and come back to this in morning :)

rmader:
Ok so the issue with the poll reloading is actually not the template - we do, like a+, successfully call into `updateDashboard` (https://github.com/liqd/adhocracy4/blob/master/adhocracy4/dashboard/assets/dashboard.js#L9). The problem appears to be that only elements with the classes `l-menu__menu` and `l-menu__aside` get updated - and apparently we're not using those classes in the opin templates

euth_wagtail/templates/a4polls/poll_dashboard.html
rmader: This commit can be dropped - the a4 template works just fine

phillimorland:
@rmader ahhhhh that makes sense

fuzzylogic2000:
Yeah, makes sense, but doesn't seem like the best class to use. Can't we have an id specially for the dashboard update or something like that?

phillimorland:
yep! i'll do that now

phillimorland:
@fuzzylogic2000 yes!

rmader:
Oh please wait another sec

rmader:
There's one nit: https://www.w3schools.com/html/html_id.asp states:
> The id attribute specifies a unique id for an HTML element (the value must be unique within the HTML document).

So if I understand that right we need to use a class instead :/

phillimorland:
@rmader i disagree with using class's for js selecting but if we are going to do this intentionally then we should decide on a naming convention before we implement to ensure they are identifiable. Camel case we already do this with react and don't we use snake case for python? what others are there? @fuzzylogic2000 @Rineee any ideas, preferences?

fuzzylogic2000:
Oh, yeah, right, an id should only be used once, so we need to use a class. :(
That's annoying because we could have kept the classes for styling and used the id for other things like that. Maybe we can use update_dashboard as the class instead then, so that at least the naming of the class makes sense and is not also used for styling?
I don't care which case we use for this. Maybe not camel. With BEM we want to use this weird snake-kebap-case mix for the classes anyway, right?

phillimorland:
I think we should use something that can't be mistaken for a styling class as it leads to confusion like we had before

fuzzylogic2000:
You're right! Is update_dashboard clear enough or should we use CamelCase then to make it totally clear?

rmader:
Maybe something like `js-selector-update-dashboard`? Naming things... :(

phillimorland:
naming things sucks! ok js-selector wins as will ensure no confusion :p

euth_wagtail/templates/a4dashboard/base_dashboard_project.html
```diff
@@ -38,7 +38,7 @@ <h2 class="dashboard-left-nav-header">{% trans 'Module'%}: {{ module_menu.module
             {% block dashboard_project_content %}{% endblock %}
         </div>
 
-        <div class="col-md-3">
+        <div id="update_dashboard" class="col-md-3">
```
rmader, line 41: This was left out ;)

rmader: Took the freedom to fix that

rmader:
Awesome!

euth_wagtail/templates/a4dashboard/base_dashboard_project.html
phillimorland: thank you! :)

PR #1852, opened 2020-04-28
Some cleanups

rmader:
Test were failing because apparently there's new timezone ("Nuuk") :)

Also, there're not enough forks of leaflet markercluster yet, only 813


fuzzylogic2000:
And why didn't pyup fail, then? Anyway!

rmader:
> And why didn't pyup fail, then? Anyway!

Could well be that timezones are directly fetched from the OS - so kinda behind our backs :)

And apparently the same happened a year ago - when "America/Godthab" was removed....

PR #1855, opened 2020-04-28
Ks 2020 04 add image to topic

Rineee:


euth_wagtail/settings/base.py
```diff
@@ -318,6 +318,7 @@
     'avatar': {'min_resolution': (200, 200)},
     'idea_image': {'min_resolution': (800, 200)},
     'tileimage': {'min_resolution': (500, 300)},
+    'topic_image': {'min_resolution': (800, 200)},
```
fuzzylogic2000, line 321: As long as they don't differ, you can use the same one as for the idea, IMHO.

euth_wagtail/settings/base.py
```diff
@@ -332,6 +333,7 @@
         'avatar_small': {'size': (60, 60), 'crop': 'smart'},
         'org_avatar_small': {'size': (60, 60), 'crop': 'scale'},
         'org_avatar_medium': {'size': (200, 200), 'crop': 'scale'},
+        'topic_image': {'size': (800, 0), 'crop': 'scale'},
```
fuzzylogic2000, line 336: Same here!

euth/communitydebate/templates/euth_communitydebate/topic_form.html
```diff
@@ -55,6 +55,18 @@ <h3 class="sans-serif">
                 {% endif %}
                 {{ form.description.errors }}
             </div>
+
+            <div class="form-group">
+                <div class="input-group form-control-upload row">
+                    <div class="col-sm-8 col-md-9">
+                        <label>Upload a picture</label>
+                        <p class="text-warning">Please make sure the image is at least 800x200 px in dimensions and at most 5 MB in file size.</p>
```
fuzzylogic2000, line 63: Even though it's the same in the idea form, I think this is not the proper way to do it. No need to fix that in this PR, though. It's not even translated? Weird!

euth_wagtail/settings/base.py
Rineee: Yes, I thought about that and decided its clearer/less confusing to have its own..but alright, I will delete it then!

Rineee: Haha, I was lying.. now when changing it I remember that also having to change the model again made me decide to use its own :)

fuzzylogic2000:
I am confused. I strated reviewing a while ago and thought, I hadn't added that, but now you already fixed it?! I will use morning brain to look at it again!

fuzzylogic2000:
@phillimorland @rine used the same one as in the ideas. I also thought it's all to big, when I tried it out, but should be the same as in ideas. It's all a bit weird, but I think, I will merge and we can sort it out together with the ideas.

PR #1886, opened 2020-05-11
Ks 2020 05 add fileupload forms dynamically

Rineee:
Need to merge #1882 before.

Todo: when reloading the page, unsaved forms are not shown anymore, but counted in TOTAL_FORMS, thus no more new forms can be added


euth/communitydebate/templates/euth_communitydebate/topic_form.html
```diff
@@ -1,7 +1,14 @@
 {% extends "base.html" %}
-{% load widget_tweaks i18n thumbnail base_tags form_tags %}
+{% load static widget_tweaks i18n thumbnail base_tags form_tags %}
+
+{% block title %}
+{% if mode == 'create' %}
+{% trans 'Create a new topic' %}
+{% elif mode == 'update' %}
```
fuzzylogic2000, line 7: Good idea! One could do with {% else %}, though, so that there is always something there (and it's one check less).

PR #2012, opened 2020-09-15
Pm 2020 09 break word

### LINKED ISSUE #2008: Particpate org tiles overlap
**URL:** https://opin.me/en/orgs/
**user:** all
**expected behaviour:** no overlap

Screenshot?

phillimorland:
also quick fix #2008 still little off on ipad but better then before


fuzzylogic2000 (approved):
Very nice!

PR #2040, opened 2020-10-12
ideas/views: return 404 instead of server error if module does not ex…

Rineee:
…ist anymore

hopefully fixes https://sentry.liqd.net/sentry/opin-prod/issues/750/events/37858/

### DIFF
--- a/euth/ideas/views.py
+++ b/euth/ideas/views.py
@@ -1,4 +1,5 @@
 from django.contrib import messages
+from django.shortcuts import get_object_or_404
 from django.urls import reverse
 from django.utils.translation import ugettext as _
 from django.views import generic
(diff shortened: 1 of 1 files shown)


fuzzylogic2000 (commented):
It looks good. But how can we not have the module? It needs to be there and should be wherever we click the button to add an idea. :thinking: Was it not published or something like that?

fuzzylogic2000:
And why is self a whole module query set? Is this a problem from multi modules or something like that? This is all a bit weird.

Rineee:
> It looks good. But how can we not have the module? It needs to be there and should be wherever we click the button to add an idea. thinking Was it not published or something like that?

I think we can only not have the module if it is really not there anymore, so if the project was deleted. Couldn't it be that people send the link around? 
I am also not sure if this is it, but I figured that returning the get_object_or_404() instead of just get() makes sense anyway and maybe its worth a try ..

Rineee:
> And why is self a whole module query set? Is this a problem from multi modules or something like that? This is all a bit weird.

Which self and where?

fuzzylogic2000 (approved):
Alright! Talking about it this seems like the only option!

PR #2055, opened 2020-10-15
euth/dashboard//organisations: make switching of tabs of multilanguag…

Rineee:
…e organisation information form work

So the changing tabs is working now, but on page reload the last active tab-panel is somehow cached and activated. This was a problem before, so maybe we can also just leave it as an issue...

@philli If you have time and do something there, could you add it as a new commit so I can see the changes and learn? Thank youuuu!  

Oh, and there was a linting error, I could not handle this quickly, so committed anyway. Is that a hint maybe?  
139:23  error  Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead.

### DIFF
--- a/euth/blueprints/templates/euth_blueprints/includes/step-aim.html
+++ b/euth/blueprints/templates/euth_blueprints/includes/step-aim.html
@@ -7,7 +7,7 @@
         </a>
         <div class="dst-lightbox-info float-right">{% trans 'Decision support tool' %}</div>
     </div>
-    <h2 class="h3">{% trans 'What do you want to do in your participation project?' %}</h2>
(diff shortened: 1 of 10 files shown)


phillimorland:
@Rineee boom baby!!! you were totally right with lint error, reloads remember page anchors so swapping link tag to button totally worked, I also did a couple of style updates that I found when clicking around that are explained in commits, i'll leave this for you to look at and merge :)

phillimorland (approved):
👍

Rineee:
> @Rineee boom baby!!! you were totally right with lint error, reloads remember page anchors so swapping link tag to button totally worked, I also did a couple of style updates that I found when clicking around that are explained in commits, i'll leave this for you to look at and merge :)

Wohoo, so cool, thank you!

PR #2056, opened 2020-10-16
communitydebate/views: initialize upload forms when form is not valid

Rineee:
fixes https://sentry.liqd.net/sentry/opin-prod/issues/1198/events/

I think we have contradictory linters now? When trying to commit, I got:

`/home/katharinaschmid/a4-opin/euth/documents/static/documents/Paragraph.jsx`
`119:19  error  Expected indentation of 20 space characters but found 18  react/jsx-indent-props`

So I changed that and then got:

`/home/katharinaschmid/a4-opin/euth/documents/static/documents/Paragraph.jsx`
`119:1  error  Expected indentation of 18 spaces but found 20  indent`

### DIFF
--- a/euth/communitydebate/views.py
+++ b/euth/communitydebate/views.py
@@ -97,6 +97,9 @@ class TopicCreateView(PermissionRequiredMixin, generic.CreateView):
                                      _('Topic '
                                        'successfully created'))
                 return response
+
+        upload_forms = forms.TopicFileUploadFormset(request.POST,
+                                                    request.FILES)
         return render(request, self.template_name,
                       self.get_context_data(upload_forms=upload_forms))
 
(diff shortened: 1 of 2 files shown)


phillimorland:
@Rineee when adding document file but no title then both form field errors are triggered instead of just the title error, doesn't do it other way around, is this to do with this?
Oh and linting errors should be fixed after config merge and rebase

Rineee:
> @Rineee when adding document file but no title then both form field errors are triggered instead of just the title error, doesn't do it other way around, is this to do with this?
> Oh and linting errors should be fixed after config merge and rebase

No, this doesnt have to do with this and must have been there before, which doesnt really make it better... will look into it!

phillimorland:
will add issue for the image error message

PR #2146, opened 2021-04-23
deps: update a4 and adapt py code to access enum

fuzzylogic2000:
The first commit only fixes the py code in so far that the tests succeed and the views are alright. 
But we still need to think about what to do with the access enum. If we need to use all three, we also need to make sure the semipublic makes sense for the projects. And I think as the enum can't be overwritten, we have to use all three (or finally change from enum to choices, which is too late now).
Also, the js code is currently broken. :grimacing:


Rineee (approved):
Yes! Let us go on from here :)

PR #2162, opened 2021-04-29
assets/scss: fixing poll button #2106

### LINKED ISSUE #2106: Show Preresults in Poll is broken
**URL:https://opin-stage.liqd.net/de/projects/christmas/** 
**user:Initiator** 
**expected behaviour:If I set up a poll and publish it there is the Text: "Vorläufiges Ergebniss anzeigen" and I can click that and see results without voting** 
**behaviour:If I set up a poll there is the Text: "Vorläufiges Ergebnisse anzeigen" but it is not a Button or any field to do this,just the text. So I have to vote to see results or wait till the poll is over. But as Initiator I maybe don't want to vote, just see the PreResult.** 
**important screensize:**
**device & browser: Firefox82.0.2 (64-Bit), Safari Version 14.0 ** 
**Comment/Question:** 

Screenshot?

khamui:
Adding btn class to style as a link. Adjusting
Poll options to be listed vertically.
Fixes #2106

### DIFF
--- a/euth_wagtail/assets/scss/components/_poll.scss
+++ b/euth_wagtail/assets/scss/components/_poll.scss
@@ -4,6 +4,8 @@
 }
 
 .poll__rows {
+    display: flex;
+    flex-direction: column;
     padding-top: 1em;
     padding-bottom: 1em;
 }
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (approved):
Nice!

PR #2167, opened 2021-04-29
Ks 2021 04 make semi public projects work

Rineee:
I wanted to add some styling for the radio buttons of the choices of project access in basic settings of project, but didnt succeed..

I wanted to add something like:
`.widget-radioselect {
    ul {
        list-style: none;
        margin: 0;
        padding: 0;
    }
}
`
but didnt really know where to put it, so its used.. can you help me with that @fuzzylogic2000 or @khamui ?

Also, the helptext for access is now gone, because we removed it in A4, but I guess it is alright, because the choices are self-explanatory?

Open question: probably people should also be able to apply for membership in semi-public projects, right? Maybe we can discuss where to put that tomorrow @fuzzylogic2000 ?

### DIFF
--- a/euth/memberships/dashboard.py
+++ b/euth/memberships/dashboard.py
@@ -13,7 +13,7 @@ class MembershipRequestComponent(DashboardComponent):
     label = _('Members')
 
     def is_effective(self, project_or_module):
-        return project_or_module.is_private
+        return project_or_module.is_private or project_or_module.is_semipublic
 
     def get_base_url(self, project):
         return reverse('a4dashboard:members', kwargs={
@@ -34,7 +34,7 @@ class MembershipInvitesComponent(DashboardComponent):
     label = _('Invited Members')
 
     def is_effective(self, project_or_module):
-        return project_or_module.is_private
+        return project_or_module.is_private or project_or_module.is_semipublic
 
     def get_base_url(self, project):
         return reverse('a4dashboard:invites', kwargs={
--- a/euth/projects/overwrites.py
+++ b/euth/projects/overwrites.py
@@ -1,0 +1,14 @@
+from django.utils.translation import ugettext_lazy as _
+
+from adhocracy4.projects.enums import Access
+
+
+def overwrite_access_enum_label():
+    Access.__labels__ = {
+        Access.PRIVATE: _('Only invited users can see content and can '
+                          'participate (private).'),
+        Access.PUBLIC: _('All users can see content and can participate '
+                         '(public).'),
+        Access.SEMIPUBLIC: _('All users can see content, only invited users '
+                             'can participate (semi-public).')
+    }
--- a/euth_wagtail/assets/scss/components/_forms.scss
+++ b/euth_wagtail/assets/scss/components/_forms.scss
@@ -7,3 +7,4 @@
 @import "forms/form-section";
 @import "forms/poll-radio";
 @import "forms/questionform";
+@import "forms/widget-radioselect";
--- a/euth/memberships/templates/euth_memberships/emails/invite.en.email
+++ b/euth/memberships/templates/euth_memberships/emails/invite.en.email
@@ -9,7 +9,7 @@ You are invited to join the project “{{ invite.project.name }}” on <i>{{ sit
 {% endblock  %}
 
 {% block content %}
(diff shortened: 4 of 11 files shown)


euth/projects/overwrites.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def overwrite_access_enum_label():
+    Access.__labels__ = {
+        Access.PRIVATE: _('Only invited users can see content and can '
+                          'participate (private).'),
+        Access.PUBLIC: _('All users can see content and can participate '
+                         '(public).'),
+        Access.SEMIPUBLIC: _('All users can see content, only invited users '
+                             'can participate (semi-public).')
```
fuzzylogic2000, line 13: Maybe we can put that into a4? We tried to overwrite the enum with two options in there, but as that is not working anyway...
I started something last week, but then didn't finish because I just wanted the update to get working and then forgot about that. https://github.com/liqd/adhocracy4/pull/713 Maybe you can use it?!

fuzzylogic2000 (approved):
Nice! 
The issue with the weird radio select was already there before, so we could add an issue and fix that in the design sprint instead? But I will also have a try. :)
Screenshot from 2021-04-29 14-02-27

As for the a4 thing, we can also leave it like it is and merge. Whatever you think is better.

fuzzylogic2000:
> Open question: probably people should also be able to apply for membership in semi-public projects, right? Maybe we can discuss where to put that tomorrow @fuzzylogic2000 ?

Right! And I also thought about that. Hmm, would be nicer to have it, but is it worth the effort? We do this weird dispatch thing to add that to the private project view. But yes, let's have a look tomorrow!

khamui:
> I wanted to add some styling for the radio buttons of the choices of project access in basic settings of project, but didnt succeed..
> 
> I wanted to add something like:
> `.widget-radioselect { ul { list-style: none; margin: 0; padding: 0; } } `
> but didnt really know where to put it, so its used.. can you help me with that @fuzzylogic2000 or @khamui ?
> 
> Also, the helptext for access is now gone, because we removed it in A4, but I guess it is alright, because the choices are self-explanatory?
> 
> Open question: probably people should also be able to apply for membership in semi-public projects, right? Maybe we can discuss where to put that tomorrow @fuzzylogic2000 ?

About the `<ul>` styling, I will have a look soon.

fuzzylogic2000:
> The issue with the weird radio select was already there before, so we could add an issue and fix that in the design sprint instead? But I will also have a try. :)

Ok, yeah, the new options with longer labels make it much worse... In a+, we add it directly to to the forms component: https://github.com/liqd/adhocracy-plus/blob/d07f68428ff391c1ab3aae3018d306f56b436023/adhocracy-plus/assets/scss/_form.scss#L117

fuzzylogic2000:
@khamui I already added a comment about the list styling. :) But you can also either add it to the form component or add a new component to do that.

Rineee:
> > The issue with the weird radio select was already there before, so we could add an issue and fix that in the design sprint instead? But I will also have a try. :)
> 
> Ok, yeah, the new options with longer labels make it much worse... In a+, we add it directly to to the forms component: https://github.com/liqd/adhocracy-plus/blob/d07f68428ff391c1ab3aae3018d306f56b436023/adhocracy-plus/assets/scss/_form.scss#L117

Yes, I tried that, but then it didnt show.. but maybe then it was a problem with updating again? I'll try again..

Rineee:
Ok, so I added styling as well as I could :) Turned out that it really just didnt show, npm cache clear --force helped again.

The enum labels are different here than in Aplus and mB, so we decided to leave them in here.

For the membership, I will ask Kathi if she wants a story.

euth/memberships/dashboard.py
```diff
@@ -13,7 +13,7 @@ class MembershipRequestComponent(DashboardComponent):
     label = _('Members')
 
     def is_effective(self, project_or_module):
-        return project_or_module.is_private
+        return project_or_module.is_private or project_or_module.is_semipublic
```
fuzzylogic2000, line 16: right!

euth_wagtail/assets/scss/components/_forms.scss
```diff
@@ -7,3 +7,4 @@
 @import "forms/form-section";
 @import "forms/poll-radio";
 @import "forms/questionform";
+@import "forms/widget-radioselect";
```
fuzzylogic2000, line 10: yes, that looks like the way, we do it in opin. :)

fuzzylogic2000 (approved):
Yeah! LGTM!

PR #2172, opened 2021-04-29
image issues

fuzzylogic2000:


khamui:
looking good! both issues fixed

PR #2176, opened 2021-04-30
locale: makemessages

Rineee:
New strings as pushed to transifex. Also old (commented out) strings deleted from source files.


fuzzylogic2000 (approved):
:+1:

PR #2232, opened 2021-07-09
Kl 2021 07 update bootstrap

fuzzylogic2000:
Still to do:

- [x] check https://getbootstrap.com/docs/5.0/migration/#sass
- [x] check https://getbootstrap.com/docs/5.0/migration/#color-system
- [x] check https://getbootstrap.com/docs/5.0/migration/#grid-updates
- [x] check https://getbootstrap.com/docs/5.0/migration/#content-reboot-etc
- [x] check https://getbootstrap.com/docs/5.0/migration/#rtl (might already be done with the utility components, though?!)
- [x] check https://getbootstrap.com/docs/5.0/migration/#forms
- [x] check that all the form-groups do not change anything, even though removed from bs5, but only added a margin before (https://getbootstrap.com/docs/4.6/components/forms/#form-groups)
- [x] check if ck-edtor accordian looks like should be (styled with card-stuff)
- [x] check https://getbootstrap.com/docs/5.0/migration/#badges, especially for `euth_wagtail/templates/account/email.html`, the other badges seem to be our own styles
- [x] replace btn-block https://getbootstrap.com/docs/5.0/migration/#buttons
- [x] check navbars https://getbootstrap.com/docs/5.0/migration/#navbars
- [x] timeline popover (seems to work, but not styled?! attempt: https://github.com/liqd/a4-opin/pull/2232/commits/5c5c4f4e0db6f1c0cb0d0b995db909eaf26acdf6 and also app.js changed, which didn't help, but broke the linter https://github.com/liqd/a4-opin/pull/2232/commits/1b46ef6c1abf9281d2ff4b697c65ccddfe3440be NOTE: doesn't work, what is ebing seen is just title tooltip i think
- [x] check that phase date in dashboard is fine (probably with from updates!)
- [ ] ~~check idea sorting (Still has data-toggle instead data-bs-toggle)~~ A4 DEP
- [x] check wagtail
- [x] check euth_dashboard dropdown
- [x] check map_choose_polygon
- [ ] fix the language stuff in the orga form (probably jQuery problem?) and the whole orga-form. Errors are not showing (because of form-text instead form-error? even though only replacing didn't fix it) and should all fields look required? - no
- [ ] ~~variables file~~ - To be completed in other pull request
- [ ] ~~image (and probably also file upload) not aligned, because row had negative margin before~~ - Can't reproduce
- [ ] ~~Comments in a4 need updating~~ A4 DEP


fuzzylogic2000:
We do [not display .form-control-file](https://github.com/liqd/a4-opin/blob/3341219a2c470c742e12d92cc6f5d8e2add126ca/euth_wagtail/assets/scss/components/forms/_form-control-upload.scss#L14), but use it in a few places. It is dropped by bootsrap5 (https://getbootstrap.com/docs/5.0/migration/#forms), but when deleting the class, The [style from bs4](https://getbootstrap.com/docs/4.6/components/forms/#form-controls) returns. Needs to be checked.

fuzzylogic2000:
ck-editor accordion currently broken on dev, so can't check if still same style. But works again! :tada:

phillimorland:
I will go one with fixing the org form but this can already be merged I think

Rineee (approved):
Its looking good! So I will be brave and merge :muscle:

PR #2234, opened 2021-07-12
bootstrap: update variables file

fuzzylogic2000:
Took the first version of the isntalled bootstrap varoiables file from https://github.com/liqd/a4-opin/commit/906118b9eabfb4e9c7432cf6b5552f80a77f7aec
and compared it to our current variables file

### DIFF
--- a/euth_wagtail/assets/scss/components/_user-indicator.scss
+++ b/euth_wagtail/assets/scss/components/_user-indicator.scss
@@ -86,7 +86,7 @@
     .user-indicator .btngroup > a {
         display: block;
         height: 50px;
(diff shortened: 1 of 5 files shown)


fuzzylogic2000:
@phillimorland @Rineee I think this is good to go! I removed a lot and had to add two variables back in again, but I think, now it is looking alright! I didn't even see a difference in these grays, or it didn't spring out and might just be fine like that.

fuzzylogic2000:
Oh, I still need to fix the colour for the linter. :( Or can I just tell it to ignore that?

phillimorland (approved):
Nice

PR #2236, opened 2021-07-13
Ks 2021 07 update a4

Rineee:
So it is working, but styling pretty broken.. I copied styles from mB and commented out any non-present variables as I heard @fuzzylogic2000 is working on them anyway?


fuzzylogic2000:
I fixed the linting errors by making comments out of the old styles. Now it looks even more broken. How do we go on with this? Do we merge and then work on the styling?

phillimorland:
closed in favour of https://github.com/liqd/a4-opin/pull/2243

PR #2237, opened 2021-07-13
WIP Pm 2021 07 bs update fixes org form

phillimorland:


phillimorland:
@Rineee so it's not working and I think I have added everything, I found your pull request but this seems to have been to make it work with straings right? 
https://github.com/liqd/adhocracy-plus/pull/822

PR #2242, opened 2021-07-15
deps/python: update wagtail to 2.11 lts

fuzzylogic2000:
This wasn't too bad. And it is fine on a fresh DB.
But I did not have a lot of wagtail in my local not-fresh DB. Could you also test it on yours @phillimorland @Rineee ? I will also add more wagtail to my unupdated DB and check it out more thoroughly after lunch.


fuzzylogic2000:
Alright, I added more wagtail, updated and it looked fine! I think this is ready.

Rineee (approved):
Looks all fine!

PR #2254, opened 2021-07-20
Pm 2021 07 use cases

phillimorland:


home/models/blocks.py
```diff
@@ -101,10 +125,72 @@ class ColumnBlock(core_blocks.StructBlock):
...
+        icon = 'grip'
+        label = '1-3 Column CTA Block'
+
+
+# 1- 2 column block with an optional bg colour and CTA
+class HighlitedColumnBlock(core_blocks.StructBlock):
+    column_count = core_blocks.ChoiceBlock(choices=[
+        ('6', 'two columns'),
+        ('12', 'one column'),
+    ], icon='cup', required=False, help_text='Only add this number of columns')
```
Rineee, line 185: Nice with the helptext and the column choice on top!

home/templates/home/blocks/highlighted_column_block.html
```diff
@@ -0,0 +1,85 @@
+{% load i18n wagtailcore_tags %}
+
+<!-- single column template checks to ensure doesn't dispay additional columns -->
+{% if self.column_count == "12" %}
+  {% for item in self.column %}
+    {% if forloop.counter <= 2 %}
```
Rineee, line 6: you need <2 or <=1 here

home/templates/home/blocks/highlighted_column_block.html
```diff
@@ -0,0 +1,85 @@
...
+        </div>
+      </div>
+    {% endif %}
+  {% endfor %}
+<!-- 2 column template -->
+{% else %}
+  <div class="container">
+    <div class="row justify-content-between">
+      {% for item in self.column %}
+        {% if forloop.counter <= 3 %}
```
Rineee, line 47: and then here <=2 or <3

Rineee (approved):
One little thing with the checking number of columns, but otherwise really cool and working great!

home/templates/home/blocks/highlighted_column_block.html
phillimorland: Good point! although, after testing I will actully use splice in the forloop instead of this method

Rineee: Ok, so I merge it like this and you change it in the other branch or whats the plan?

PR #2255, opened 2021-07-20
register/howto: first attempt using carousel

khamui:


euth_wagtail/templates/account/signup.html
```diff
@@ -5,109 +5,114 @@
 {% block title %}{% trans 'Register an account' %}{% endblock %}
 
 {% block content %}
```
khamui, line 7: except of the last lines in this file, its just indentation.

euth_wagtail/assets/scss/components/_howto.scss
```diff
@@ -0,0 +1,110 @@
+.howto__container {
```
khamui, line 1: i introduced this file, because i think it is a lot of very specific styles for the howto screen.
what do you think @phillimorland ?

euth/contrib/templatetags/contrib_tags.py
```diff
@@ -24,3 +25,39 @@ def limited_paginator(page_obj):
         current_index < last_index - ADDITIONAL_PAGES else last_index
 
     return page_obj.paginator.page_range[start_index:end_index]
+
```
khamui, line 28: most probably this is the wrong spot. i would like to extract the data for the steps of the howto screen, but no idea how to make it available properly so i misused `contrib_tags` for now. @Rineee maybe you have a better location or idea to organize this?

phillimorland: I would say as the contents won't change we can just put it directly into the carousel in the template, what do you think @rine?

euth_wagtail/assets/js/euth_wagtail.js
```diff
@@ -87,6 +87,15 @@ $(document).ready(function () {
...
+
+  $('.howto-carousel').slick({
+    dots: true,
+    infinite: false,
+    arrows: false,
+    cssEase: 'linear',
+    autoplay: true,
+    autoplaySpeed: 6000
+  })
 })
```
phillimorland, line 99: nice!

euth_wagtail/assets/scss/components/_howto.scss
```diff
@@ -0,0 +1,110 @@
...
+    display: block;
+    margin-left: auto;
+    margin-right: auto;
+    width: 70%;
+    max-width: 300px;
+}
+
+// slick dots overwrites
+ul.slick-dots {
+    margin: 0;
```
phillimorland, line 56: any slick overwrites should be nested under custom class name of the specific slick carousel to ensure no styles leak to other slicks on the site

euth_wagtail/assets/scss/core/_bootstrap-variables.scss
```diff
@@ -7,6 +7,7 @@ $gray-dark:             #2b2b2b;
 $gray:                  #818181;
 $gray-light:            #d8d8d8;
 $gray-lighter:          #f4f4f4;
+$bright:                #ffffff;
```
phillimorland, line 10: for white you can always use $body-bg, it comes with bootstrap, it's not there because we don't need to overwrite it so theres no way to know unless you know bootstrap :grimacing:

euth_wagtail/templates/account/howto_item.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n static thumbnail %}
+
+<div class="howto-item__container">
+  <h1>{{ howto_item.header }}</h1>
+  <img class="howto-item__image" alt="how to register image" src="{% static howto_item.img %}">
```
phillimorland, line 5: for a11y reasons, there should only be a single h1 element on each page, in this case I think it should be 'Registration' or 'Login' the slide headings can be h2 or h3, i'm never quite sure in such a mixed use page, whatever looks better is sometimes the best option

euth_wagtail/assets/scss/components/_howto.scss
phillimorland: totally agree!

euth_wagtail/templates/account/signup.html
```diff
@@ -5,109 +5,114 @@
...
-                {% include "socialaccount/snippets/provider_list.html" with process="login" %}
-            </ul>
-        </div>
-        {% endif %}
     </div>
-    {% endblock %}
-</div>
+    <div class="account__container--right">
+        {% include "./howto.html" %}
+    </div>
```
phillimorland, line 115: I think the django way is to always add includes into the includes folder @Rineee is that right?

phillimorland (changes requested):
supe cool and totally working! :D theres a few django best practices questions for Rine and some css comments. 
The design also shows new registration although i'm not sure how we are meant to do that as we don't have a way to click between registration and login so I will ask in the story about that.

euth_wagtail/templates/account/signup.html
Rineee: Yes, thats right! And better not use ./, but specify the path with the whole folder structure after /templates, so if you put it into euth_wagtail/templates/account/includes, you would need 'account/includes/howto.html'. For includes, Django will look into all folders called template and look for the path.

euth/contrib/templatetags/contrib_tags.py
Rineee: I think I would add it to the form as a property as its done here with the formsections https://github.com/liqd/a4-opin/blob/9fe7f4f99bad8814415dde96873b43d4069eaaf5/euth/accounts/forms.py#L24-48
The form is here https://github.com/liqd/a4-opin/blob/9fe7f4f99bad8814415dde96873b43d4069eaaf5/euth/users/forms.py#L9
Then when you include the howto.hmtl, you can pass the howto_items likes this: {% include "howto.html" with howto_items=form.howto_items %}

euth_wagtail/templates/account/howto.html
```diff
@@ -0,0 +1,14 @@
+{% load contrib_tags %}
+<div class="howto__container">
+  <div class="howto-carousel__container">
+    <div class="howto-carousel">
+      {% get_howto_items as howto_items %}
+      {% for hi in howto_items %}
+      {% include './howto_item.html' with howto_item=hi %}
```
Rineee, line 7: Also here better put it into /includes and specify the path after templates

euth_wagtail/assets/scss/components/_howto.scss
khamui: true 🤝

euth_wagtail/templates/account/signup.html
khamui: cool! thanks for explanation

euth/contrib/templatetags/contrib_tags.py
khamui: Nice! just saw that after i pushed the last commit. I will have a look on that with the current subtask and maybe change it. but the current version (directly in carousel) works also and is okay to be merged i think.

phillimorland (approved):
Perfect! :)

PR #2256, opened 2021-07-21
Ks 2021 07 issue fixing

Rineee:


euth_wagtail/assets/scss/components/_label.scss
```diff
@@ -38,6 +38,7 @@
 
 .badge-secondary {
     // @include label-variant($brand-brand2;
+    background-color: $gray-light;
     color: $black;
```
phillimorland, line 42: just to check is there a reason why $brand-brand2 isn't used as bg colour? whenever 'secondary' is used it normally relates to the second brand colour, 'muted' may refer to gray.

phillimorland (approved):
Just one question but if brand colour didn't work then feel free to merge :)

euth_wagtail/assets/scss/components/_label.scss
Rineee: In the issue the background was dark grey and Kathi wanted it lighter, thats why I used light grey.. should I use $brand-brand2 instead?

PR #2257, opened 2021-07-21
Pm 2021 07 story styling fixes

phillimorland:


home/templates/home/blocks/highlighted_column_block.html
```diff
@@ -2,67 +2,58 @@
...
+    {% else %}
+      <div class="m-cover">
+    {% endif %}
+      <div class="container">
+        <div class="row">
+        {% if item.highlight == "highlight" %}
+          <div class="block col-12 m-highlight">
+        {% elif item.highlight == "highlight-purple" %}
+          <div class="block col-12 m-highlight-purple">
+        {% else %}
```
Rineee, line 19: You forgot the green here! And when I dont select a highlighting color, it doesnt show at all (only for 1 column). And, the link on a blue highlighted part becomes orange, is that how its supposed to be?

phillimorland: hahahahs nope, will take a look :)

Rineee:
Now the styling for one column highlighted green is broken :grimacing: 
And when no highlight is selected, the links dont show, probably because they are white on white?

phillimorland:
sorry dude! that 'should be fixed by now, I think i've checked evry column variation

Rineee (approved):
Cool, all good now!

PR #2264, opened 2021-07-22
deps: update a4 and use deps from there

### LINKED ISSUE #2245: #4410 endless loading spinner when I resubmit my answer
**URL:** 
https://opin-dev.liqd.net/de/projects/test-poll-187/
**user:** 
admin
**expected behaviour:** 
When I resubmit my answer to a poll without changing anything (i.e. after I reload the page or if I come back to the page later), the message „answer saved“ should appear, the loading spinner should disappear and the link to the preliminary results should be visible again.
**behaviour:** 
There is no message when I click on submit and the loading spinner appears but never goes away and the preliminary results link is never visible again.
**important screensize:**
all sizes
**device & browser:** 
macOS 11.4., Safari 14.1 + Firefox 89
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2021-07-18 um 15 23 04

### LINKED ISSUE #2248: Should not need to install popper
Popper is installed in a4 for poll, we should not need to install it again in Opin

Rineee:
The A4 update fixes #2245

Also now fixes  #2248


package.json
```diff
@@ -11,7 +11,7 @@
     "@babel/runtime": "7.14.0",
     "@fortawesome/fontawesome-free": "5.15.3",
     "@popperjs/core": "2.9.2",
```
Rineee, line 13: Also tried to remove that, but that gave me an error.. do you have an idea why?

phillimorland: no?! Bootstrap has it's own popper dependency so we shouldn't need this?! will have a little poke

phillimorland: ok added popper to a4 once merged we can update this hash and rm it from here

PR #2265, opened 2021-07-22
Kt 2021 07 register form adjust & howto for login screen

khamui:


phillimorland (approved):
cool!

euth_wagtail/assets/scss/components/forms/_general-form.scss
```diff
@@ -78,6 +78,23 @@
             content: "\f05A "; //info-circle
         }
     }
+
+    header {
+        &.form-inner {
+            padding-top: 6rem;
```
phillimorland, line 84: I know Opin in inconsistent but for paddings and margins we would like to always use the $spacer variable so if we ever change the size, all spaceings are adjusted, not just some.

euth_wagtail/assets/scss/components/forms/_general-form.scss
```diff
@@ -78,6 +78,23 @@
...
+        }
+
+        & > h1 {
+            font-size: 24px;
```
phillimorland, line 88: font sizes should also be a variable, I think 24px is $font-size-xlarge

euth_wagtail/templates/account/signup.html
```diff
@@ -1,22 +1,24 @@
...
+                    {% trans "I already have an account, " %}
+                    <a href="{% url 'account_login' %}?next={{redirect_field_value|urlencode}}">
+                        {% trans "login" %}
+                    </a>
+                </p>
             </header>
-
             {% block form %}
-            <form class="skinny-form-inner" role="form" name="RegisterForm" action="{% url 'account_signup' %}" method="post" >
+            <form class="form-inner" role="form" name="RegisterForm" action="{% url 'account_signup' %}" method="post" >
```
phillimorland, line 21: the header and form are both using form-inner which is creating too much padding between header and form

phillimorland (changes requested):
Just a couple of bits but looking good and mobile too :) also the images for the steps are in `/Nextcloud/Projekte/OPIN (EUth, DigY)/_Kreation/210720-registration-steps` so they could also be added

phillimorland (approved):
very nice!

PR #2273, opened 2021-07-27
Pm 2021 07 comment fixes

phillimorland:


phillimorland:
this got a little out of hand, once again using px spacings due to it already being standard for many things :(

home/templates/home/blocks/tile_column_block.html
```diff
@@ -1,110 +1,220 @@
...
+      </div>
+      {% endif %}
+    </div>
 
-          {% if self.image_col1 %}
-            {% image self.image_col1 fill-300x268 class="card-img-top" %}
-          {% else %}
-            <img src="{% static 'images/yellow-bg.png'%}" class="card-img-top" />
-          {% endif %}
+<!-- Mobile version -->
```
phillimorland, line 17: @fuzzylogic2000 this is the looping issue

home/templates/home/blocks/tile_column_block.html
```diff
@@ -1,110 +1,220 @@
...
-          {% if self.image_col2 %}
-            {% image self.image_col2 fill-300x268 class="card-img-top" %}
-          {% else %}
-            <img src="{% static 'images/light-blue-bg.png'%}" class="card-img-top" />
+<!-- Desktop version -->
+      <div class="row d-none d-lg-flex">
+        <div class="col-lg-3 d-flex align-items-stretch">
+          <div class="card tile-block__tile">
+          {% if self.internal_link_col1 %}
+            <a href="{{ self.internal_link_col1.url }}">
```
phillimorland, line 118: @fuzzylogic2000 maybe here is better, all the cols are number to ensure notmore than 4 are added but becasue of that I couldn't figure out how to loop through, instead i'm duplicating code whish I don't like

fuzzylogic2000 (approved):
Cool, it looks good! 
And I am fine with leaving the columns like that. If we ever need to they will be easy to migrate.

PR #2279, opened 2021-07-29
dashboard/language_switch: add preventDefault to stop page reload and…

### LINKED ISSUE #2249: organisation form broken
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** the form to add description etc in different languages is broken
**important screensize:**
**device & browser:** 
**Comment/Question:** 
Form is broken since this commit https://github.com/liqd/a4-opin/commit/c4818ce91a1fac016223ea6d41a751a21b905059 , but going back to react 16.x does not help. Updating Aplus to react 17 also breaks the form there.

The page reloads on click of a language button (can be removed by adding preventDefault inside react_language_switch.activateTab) but the activation of  the corresponding tab content still does not work (stays on engl content.)

The onClick event is missing for the broken form, see screenshots.
button_broken
button_working

phillimorland:
… add required data attributes and classes for tabs to work again, adapt styling fixes #2249

### DIFF
--- a/euth/dashboard/static/language_switch/react_language_switch.jsx
+++ b/euth/dashboard/static/language_switch/react_language_switch.jsx
@@ -35,9 +35,10 @@ class LanguageSwitch extends React.Component {
     }
   }
 
-  activateTab (e) {
+  handleActivateTab (e) {
     const languageCode = e.target.textContent
     this.setState({ activeTab: languageCode })
+    e.preventDefault()
   }
 
   addLanguage (e) {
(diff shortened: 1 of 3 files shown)


fuzzylogic2000 (approved):
Soooo cool!
There is the tiny thing, that if you delete an activated language, it still shows the removed language until it is saved. But as it is currently really broken and noone noticed, I think we don't care! I mean, I don't. :)

Rineee:
@phillimorland Yeah, so cool! So if I get it right, it was crucial to also add the classnames to make it work?

phillimorland:
@Rineee yeah, so there are 2 js things happening here, 1 is the switch with adding and removing the buttons which we made (which I originally thought was broken) The other is the tabing between the textareas, that functionality is provided by bootstrap js and it was triggered by elements with the correct data attributes, I think the bootstrap js code changed and it now also utilises the classnames in the tab function so it broke.

PR #2280, opened 2021-07-29
translations: make messages

phillimorland:


fuzzylogic2000:
Did you push these strings already? Then we should also merge them in order to have transifex and the locale folder in the same state.

phillimorland:
@fuzzylogic2000 ahh ok then yes it's ready, all pushed and Katha has been tagged

fuzzylogic2000 (approved):
Looking good!

PR #2298, opened 2021-08-05
settings: remove outdated maps key

goapunk:
seems to be a remnant of the flashpoll


fuzzylogic2000 (approved):
Yeah! Goodbye (more of the) flashpoll!

PR #2311, opened 2021-08-26
nav/submenu: no underline

khamui:


fuzzylogic2000 (approved):
Looks good and I guess, you also checked that all other dropdowns still look like they should? 
Maybe that's also and more a question for @hello-kathi : Are all other dropdowns already looking like they should? I also found the idea filters, that are underlined. Should they stay like that?

hello-kathi:
@fuzzylogic2000 ah thanks for looking. No we had the underlined links nowhere before, so it shouldn't be in any dropdown. I just checked the filters and there it is (already) gone.
But I found another one... I will write an new issue, sorry for missing that.
Bildschirmfoto 2021-08-30 um 12 08 12

PR #2321, opened 2021-09-01
general styles: setting all opin anchor elements '<a>' to not underlined

### LINKED ISSUE #2316: remove dropdown underlines from links (and buttons) in menu
@fuzzylogic2000 ah thanks for looking. No we had the underlined links nowhere before, so it shouldn't be in any dropdown. I just checked the filters and there it is (already) gone.
But I found another one... I will write an new issue, sorry for missing that.
Bildschirmfoto 2021-08-30 um 12 08 12

_Originally posted by @hello-kathi in https://github.com/liqd/a4-opin/issues/2311#issuecomment-908217289_

khamui:
closes #2316 

think setting all `<a>`-elements to no underline is better.

### DIFF
--- a/euth_wagtail/assets/scss/core/_type.scss
+++ b/euth_wagtail/assets/scss/core/_type.scss
@@ -13,6 +13,7 @@ html {
 
 a {
     transition: color 0.2s;
+    text-decoration: none;
 }
 
 strong {


fuzzylogic2000 (commented):
Does this remove the underline from all links? Should we do that? I think it's not accessible at all, but am not sure that we don't do it on prod already anyway.

fuzzylogic2000:
Oh my, I checked prod and couldn't find any underlines on links. :grimacing: So, I think we merge? @phillimorland What's your opinion?

khamui:
> Does this remove the underline from all links? Should we do that? I think it's not accessible at all, but am not sure that we don't do it on prod already anyway.

From the linked issue, think @hello-kathi said `No we had the underlined links nowhere before`, so in case we have underlined elements/links somewhere, those should be `underlined` explicitly?

phillimorland:
it's not accessible however Opin isn't too good with that already, we still have cursor behaviour and colour change and they are read out by SR so will merge

hello-kathi:
Thank you @phillimorland @khamui @fuzzylogic2000

PR #2349, opened 2021-09-23
home/blocks: rm requirment for headlin for video interim fix incl mig…

phillimorland:
…ration


fuzzylogic2000 (commented):
So, this needs to be merged to be deployed Monday?
This might not be interim though. Changing the field back to required is harder - we'll need a default then, but that should also be ok as we are the ones filling it?

phillimorland:
@fuzzylogic2000 so the issue is actually that the video embed in the richtext is too small and seems to be unstyled, I think this came about with the wagtail update and it would be a much longer fix, so that Kathi can add videos which are full width without having a huge headline before every video this was the solution.

PR #2352, opened 2021-09-27
templates/manuals_detail_page: add class name for styling of block ty…

phillimorland:
…pes, update styling, add classname for richtext video, rm title from video block


phillimorland:
@fuzzylogic2000 I went a bit over the top but as I was there anyway i figured I would fix the video as was easy once had the classname and also not show the blocks on the manual detail section page which are totally broken, if it's too much just tell me and i'll remove!

fuzzylogic2000 (approved):
The code looks good and as I don't have these blocks locally. I will just merge to be tried on dev?!

phillimorland:
yeah everything is on dev

PR #2408, opened 2021-11-18
Jd 2021 11 fix webpack asset module

goapunk:
also add a missing migration


fuzzylogic2000 (approved):
Right, the user model needs migrations whenever the timezones change. :)

PR #2409, opened 2021-11-18
Jd 2021 11 signup captcha

goapunk:
Add captcha to the signup form. Will probably need styling?


goapunk:
arg, accidentally merged this. just ignore it.

PR #2437, opened 2022-02-28
Ks 2022 02 adapt ci to ubuntu 2004

Rineee:
This whole part is missing in the django.yml here: https://github.com/liqd/adhocracy-plus/blob/14ff0be72ed434253448f431b20f5c240f174e07/.github/workflows/django.yml#L12 Do we also need/want that?


Rineee:
Ahh, will check/fix these tests later when Im back..

fuzzylogic2000:
> Ahh, will check/fix these tests later when Im back..

I think, you have to use different email addresses for every test: https://github.com/liqd/a4-meinberlin/commit/af2b4a37a0a2b4001519f4a3e54d2cb37ac85830

fuzzylogic2000 (approved):
Looks good to me! Maybe we can merge (when tests are fine), so that you can go on updating and @goapunk can answer the question tomorrow. :)

Rineee:
@fuzzylogic2000 It was the captcha missing in form data of tests.. now it should be fine!

Rineee:
@fuzzylogic2000 What the hell, another migration?

PR #2441, opened 2022-03-02
Jd 2022 03 transifex client migration

goapunk:
* remove deprecated transifex client
* change docs for translation
* update translations


fuzzylogic2000 (approved):
Looks good! Let's merge when the build is fixed.

PR #2445, opened 2022-03-02
Ks 2022 02 update to django 3

Rineee:
I did not update A4 to the newest version as [this commit](https://github.com/liqd/adhocracy4/commit/624297f51948db66fc77c28cdbf59fc73197e3e2) seems to break the ProjectAdminForm (I think because there are no project topics in opin). Did not have the brain so fix that tonight, but will do tomorrow and thought we could merge already, so others can go on..

Also, the wagtail migration thing was not a problem (see [here](https://github.com/liqd/adhocracy-plus/pull/1528/commits/5b16a011cd926ecfb9bc104ad55e617af1f61b1f)), i.e. its installing on both empty and filled databases or did I miss something? So I did not add that to the create_homepage migration..  but could you double check @fuzzylogic2000 that its working?

### DIFF
--- a/docs/translations.md
+++ b/docs/translations.md
@@ -2,7 +2,7 @@
 
 ## Languages
 
-  - source code uses en_GB (arguments to ugettext == msg_id )
+  - source code uses en_GB (arguments to gettext == msg_id )
      - lower case (except for first word of title or sentence)
   - support many languages with transifex (en, de, it, fr,
     sv, sl, da, el, ka, mk as of 10/2017)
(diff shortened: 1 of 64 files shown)


goapunk (approved):
Looks good to me. Don't really know about the migration, will leave that to @fuzzylogic2000

fuzzylogic2000 (approved):
Nice! 
Yes, it also works on an empty DB, so I think, we don't need to worry.

PR #2471, opened 2022-03-10
apps/comments_async: touchups

khamui:


euth_wagtail/assets/scss/core/_bootstrap-variables.scss
```diff
@@ -7,6 +7,7 @@ $gray-dark:             #2b2b2b;
 $gray:                  #818181;
 $gray-light:            #d8d8d8;
 $gray-lighter:          #f4f4f4;
+$gray-muted:            #6c757d;
```
fuzzylogic2000, line 10: Why do we need a new colour? Can't we just use the `$gray`?

fuzzylogic2000 (approved):
Nice!
I will change the new gray to one we already had in the variables, though. If there was a reason, we needed exactly that new gray, you'll have to do that in another PR. :p

euth_wagtail/assets/scss/core/_bootstrap-variables.scss
khamui: actually I was unsure here and wanted to ask, but forgot. Thanks for the catch and change @fuzzylogic2000

PR #2484, opened 2022-03-14
opin/wording: fix random variable names and emails

fuzzylogic2000:
some random wording things, I stumbled over when investigating issues.


khamui:
not really tested locally, but from reading code it looks good!

PR #2485, opened 2022-03-14
communitydebate/views: do not re validate upload forms if they were i…

### LINKED ISSUE #2058: Community debate module error messages
**expected behaviour:** Error message only shown for document upload if file is incorrect
**behaviour:** Error message is also shown when title is not added

Rineee:
…nvalid

fixes #2058

Problem was, that when the upload_form was invalid, a new upload_form instance was initialized that was then validated again (happens somewhere along the render), so that is where the second validation came from @fuzzylogic2000. And by then the file somehow gets lost, I guess this is also related to the file not being there anymore if validation fails (as for images), but I dont fully get what is happening there..

### DIFF
--- a/euth/communitydebate/views.py
+++ b/euth/communitydebate/views.py
@@ -95,8 +95,9 @@ class TopicCreateView(PermissionRequiredMixin, generic.CreateView):
                                        'successfully created'))
                 return response
 
-        upload_forms = forms.TopicFileUploadFormset(request.POST,
-                                                    request.FILES)
+        else:
+            upload_forms = forms.TopicFileUploadFormset(request.POST,
+                                                        request.FILES)
         return render(request, self.template_name,
                       self.get_context_data(upload_forms=upload_forms))
 


phillimorland (commented):
looks good, just to check the file is still lost if a file title is not added, is that expected? if so good to merge as it fixes initial issue

fuzzylogic2000:
> looks good, just to check the file is still lost if a file title is not added, is that expected? if so good to merge as it fixes initial issue

@phillimorland Yes, that is expected. That also happens for the images. :woman_shrugging:

fuzzylogic2000 (approved):
Supercool!

phillimorland:
> > looks good, just to check the file is still lost if a file title is not added, is that expected? if so good to merge as it fixes initial issue
> 
> @phillimorland Yes, that is expected. That also happens for the images. woman_shrugging

It says that in description but somehow I did not connect the dots in my head?! cheers! :)

PR #2492, opened 2022-03-15
Pm 2022 delete btn

phillimorland:
project pages were a bit random in their widths so I decided that anything non interactive (list ect.) is medium and anything interactive is narrow now this will help with UX so there's a lot of template stuff

### DIFF
--- a/euth_wagtail/assets/scss/core/_containers.scss
+++ b/euth_wagtail/assets/scss/core/_containers.scss
@@ -31,7 +31,11 @@ body {
 
 img {
     max-width: 100%;
-    height: auto;
+    height: auto !important;
+}
+
+p {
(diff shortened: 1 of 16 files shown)


euth_wagtail/assets/scss/core/_containers.scss
```diff
@@ -31,7 +31,11 @@ body {
 
 img {
     max-width: 100%;
-    height: auto;
+    height: auto !important;
+}
+
+p {
+    @include clearfix; // stops overlap of img added via ckeditor and floated
```
phillimorland, line 38: not sure how this will go but checked a bit and seemed fine

fuzzylogic2000 (approved):
Looks good!

PR #2495, opened 2022-03-16
euth/contrib: overwrite module django admin to redirect to project

### LINKED ISSUE #2136: View on site from django admin cause 500 and sentry error
**user:** admin
**expected behaviour:** clicking 'view on site' from module page in django admin takes you to page
**behaviour:** it causes a 500 error and triggers santry
**Comment/Question:** this may have an answer https://stackoverflow.com/questions/911887/how-can-i-get-the-django-admins-view-on-site-link-to-work

Sentry error: https://sentry.liqd.net/organizations/liqd/issues/1346/events/?project=8&referrer=alert_email

sabinammm:
Fixes #2136

### DIFF
--- a/euth/contrib/admin.py
+++ b/euth/contrib/admin.py
@@ -1,0 +1,15 @@
+from django.contrib import admin
+from django.urls import reverse
+
+from adhocracy4.modules import models
+from adhocracy4.modules.admin import ModuleAdmin as A4ModuleAdmin
+
+
+class ModuleAdmin(A4ModuleAdmin):
+
+    def view_on_site(self, obj):
+        return reverse('project-detail', args=[str(obj.project.slug)])
+
+
+admin.site.unregister(models.Module)
+admin.site.register(models.Module, ModuleAdmin)


euth/contrib/admin.py
```diff
@@ -0,0 +1,36 @@
...
+
+class ItemAdmin(admin.ModelAdmin):
+    list_filter = (
+        'module__project__organisation',
+        'module__project__is_archived',
+        ProjectFilter
+    )
+    list_display = ('__str__', 'creator', 'created')
+    readonly_fields = ('creator',)
+    date_hierarchy = 'created'
```
fuzzylogic2000, line 21: We don't need to copy the ItemAdmin, as this is not used in the ModuleAdmin. We use it in some projects (see mB) to display and filter the items.

euth/contrib/admin.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class ModuleAdmin(admin.ModelAdmin):
+    inlines = [
+        phase_admin.PhaseInline
+    ]
+    list_filter = ('project__organisation', 'project')
+    list_display = ('__str__', 'name')
```
fuzzylogic2000, line 29: Instead of copying this over, you could inherit from a4s ModuleAdmin and only add the view_on_site method.
```
from adhocracy4.modules.admin import ModuleAdmin as A4ModuleAdmin

class ModuleAdmin(A4ModuleAdmin):
...
```

fuzzylogic2000 (changes requested):
Works like a charm!
But there is code, we don't need to copy over.

euth/contrib/admin.py
sabinammm: I see :)

fuzzylogic2000 (approved):
Nice!

PR #2511, opened 2022-03-24
locale: translations pulled from transifex

sabinammm:
US #5530, Task #5530
@Rineee When I pulled, there were only strings from German and Slovenian, is this correct? And there is nothing else that needs to be done here?

### DIFF
--- a/locale/bg/LC_MESSAGES/django.po
+++ b/locale/bg/LC_MESSAGES/django.po
@@ -2,7 +2,7 @@
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the PACKAGE package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
-#
+# 
 #, fuzzy
 msgid ""
 msgstr ""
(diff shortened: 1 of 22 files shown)


Rineee:
> US #5530, Task #5530 @Rineee When I pulled, there were only strings from German and Slovenian, is this correct? And there is nothing else that needs to be done here?

No, I dont think so, the other languages have also new translations, that should be there.. lets look at it tomorrow!

hello-kathi:
@Rineee 
We translated all languages so there should be 10, maybe 9 if English is not counting. ;) sorry for Klugscheissern.

fuzzylogic2000 (approved):
Looks good (I only looked at the first three languages, though...)

Rineee:
@sabinammm Sorry, this is a bit late and also not super important, but for next time: we normally also do 'make po' after translations are pulled, because the formatting is a bit different then (only one code reference in one line). So this way, we always have the same format committed.. in the instructions it is [here](https://github.com/liqd/a4-opin/blob/68c70df052005ea57dda95ad739763289e64b00e/docs/translations.md?plain=1#L55), not very explicit though. Better in aplus [here](https://github.com/liqd/adhocracy-plus/blob/main/docs/languages_and_translations.md?plain=1#L33-L37). If you have time, maybe you could update that in opin? Then you could also add the force pulling.. thanks!

PR #2512, opened 2022-03-25
[#5530] translations: format pulled translations

sabinammm:
US #5530

In reference to https://github.com/liqd/a4-opin/pull/2511#issuecomment-1078872588

### DIFF
--- a/docs/translations.md
+++ b/docs/translations.md
@@ -50,10 +50,13 @@
   - required if translators worked on transifex
   - update all languages except en_GB
 
-        tx pull -a
+        tx pull -a -f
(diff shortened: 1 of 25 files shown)


Rineee:
Cool, thank you! And you pushed the new source strings to transifex now, right? So this is the file as pushed to transifex?

sabinammm:
> Cool, thank you! And you pushed the new source strings to transifex now, right? So this is the file as pushed to transifex?

Just pushed :)

Rineee (approved):
Cool, thanks!

PR #2572, opened 2022-06-14
merge AFTER A4 Update - euth_wagtail/urls: remove QuestionRouter and VoteViewSet after refact…

Rineee:
…or in A4


Rineee:
@phillimorland You are doing the A4 update, right? Are we updating to recent A4 (sorry, missed this)? 

If we do, reminder that we then also need to merge this, as QuestionRouter and VoteViewSet have been removed in A4.

phillimorland:
@Rineee I don't think we are doing the most recent one?! also may have missed that, I just updated to the a4 tag that has the react update in and was going to try wagtail if that passes tests, maybe we talk in standup

Rineee:
> @Rineee I don't think we are doing the most recent one?! also may have missed that, I just updated to the a4 tag that has the react update in and was going to try wagtail if that passes tests, maybe we talk in standup

Ah yes, just found your PR, in mB-v2206.1 version, they have not been removed yet, so dont need to merge this then..

fuzzylogic2000:
@philli-m The python installs, but the npm install fails because of `Module not found: Error: Can't resolve 'react'` in a lot of places. Do you go on in here?

philli-m:
sweet will take over, thank you! :)

fuzzylogic2000 (approved):
Supercool!

PR #2588, opened 2022-11-03
makefile: does it build with --legacy-peer-deps

fuzzylogic2000:
Doesn't work locally, but it builds, so we can merge and add it to the release?


goapunk (approved):
looks good to me, installing also worked locally for me (make install)

goapunk:
@fuzzylogic2000 ci fails because of the changed timezone information, do we care? should we add it?

fuzzylogic2000:
> @fuzzylogic2000 ci fails because of the changed timezone information, do we care? should we add it?

@goapunk I tried to last week: after installing locally, makemigrations didn't add migrations, though. So, I didn't know what to do there.

goapunk:
> > @fuzzylogic2000 ci fails because of the changed timezone information, do we care? should we add it?
> 
> @goapunk I tried to last week: after installing locally, makemigrations didn't add migrations, though. So, I didn't know what to do there.

I pushed a migration

Issue #2590, opened 2022-12-06
stop using deprecated properties from a4


fuzzylogic2000:
still used but already removed from a4:

- [ ] projects.days_left

still used, but deprecated:

- [ ] projects.active_phase

fuzzylogic2000:
The days_left and the templatetag were fixed in #2592

PR #2592, opened 2023-02-17
apps/projects: do not use deprecated property days_left from a4 - fix…

fuzzylogic2000:
…es part of #2590

I just moved that from a4 to here instead of really fixing...

failing because of missing user timezone migrations. Should we add that in here?


phillimorland:
@fuzzylogic2000 ahhh that explains some things, I think adding it here in a separate commit is fine right?

fuzzylogic2000:
> @fuzzylogic2000 ahhh that explains some things, I think adding it here in a separate commit is fine right?

@phillimorland Done!

phillimorland (approved):
sweet!

PR #2609, opened 2023-02-21
[#6814] wagtail v4 LTS update

philli-m:


euth_wagtail/settings/base.py
```diff
@@ -365,6 +365,10 @@
 
 WAGTAIL_SITE_NAME = "euth_wagtail"
 
+# Base URL to use when referring to full URLs within the Wagtail admin backend -
+# e.g. in notification emails. Don't include '/admin' or a trailing slash
+WAGTAILADMIN_BASE_URL = "http://localhost:8000"
+
```
philli-m, line 371: @goapunk I added this and wasn't sure if it also meant we would have to overwrite in our pillar but I assume not?

fuzzylogic2000: Yes, we do! Will add that as a task.

fuzzylogic2000:
Sorry, I merged some updates and produced conflicts. 
Is this ready to review apart from that?

renovate.json
```diff
@@ -28,6 +28,10 @@
       {
         "matchPackageNames": ["Django"],
         "allowedVersions": "=3.2"
+      },
+      {
+        "matchPackageNames": ["wagtail"],
+        "allowedVersions": "<4.2"
```
fuzzylogic2000, line 34: I think we decided to go with the `=` as above, so it should be `=4.1`. But also that's not suuperimportant to me.

fuzzylogic2000 (approved):
Looks good. One comment, but just merge if you don't care. :)

PR #2627, opened 2023-02-22
translations: local: add strings for translations

philli-m:


fuzzylogic2000 (commented):
Are these the new strings that are on transifex now? I think we usually just merged them and then made a new PR with the actual translations. So, feel free to merge if that's the case.

PR #2674, opened 2024-01-10
[7942] disable registration

m4ra:
testing instructions:
https://taiga.liqd.net/project/liqd-liquid-software/us/7942?milestone=386

for google signin with new google account need to check once we merge to dev. But with the signup disabled in the `euth/users/adapter.py` it should not allow new google singups


m4ra:
> @m4ra it looks great thanks! are the disables only temporary, that's why it's commented out?

they probably would want the registration back when they move to a wagtail only website.

PR #2689, opened 2024-09-12
WIP: [8306] delete apps without models and prepare deletion of apps with models

goapunk:
start removing apps without models and generate model deletion migrations for apps with models.

needs to be merged and deployed on all servers before merging https://github.com/liqd/a4-opin/pull/2690


goapunk:
@m4ra tagging you for review but let's not merge this until we archived the current opin.me

m4ra (approved):
LGTM

PR #2690, opened 2024-09-12
[8306] delete more apps and drop a4 tables

goapunk:
- delete all apps which had migrations deleting their models in the previous PR
- manual migration to drop all a4 tables
- prepare removal of organisation app
- reset migrations for home app as they depend on other migrations (see commit on how to fix migration state manually)

**This PR needs to be merged after https://github.com/liqd/a4-opin/pull/2689 was deployed!**


euth/contrib/migrations/0001_delete_a4_models.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql=[
+                "DROP TABLE IF EXISTS background_task",
```
m4ra, line 14: hmm this didn't work for background_task table. Did it work for you @goapunk?

goapunk: will double check, I copied it from mb/a+ though so i assume it should have worked

goapunk: ah, I forgot to remove the package from the installed apps and dependencies

goapunk:
@m4ra this should hopefully be ready again

m4ra (approved):
thanks!

PR #2691, opened 2024-09-12
[8306] more cleanup

goapunk:
cleanup up some more stuff which is not needed anymore.

**This PR needs to be merged after https://github.com/liqd/a4-opin/pull/2690 was deployed!**


goapunk:
actually I think this is still WIP and needs some work, let me check

m4ra (approved):
thanks

